PR #17, opened 2017-06-19
support resource identifier as a parameter "resource"

rahulrulez:
Hi Damien,

First of all, great work there with the library. I was looking to build exactly what you've built. However, I found one of the configurations missing.

Some of the requests do require you to include resource you are requesting. e.g. I have created two Azure AD apps, client and webapi. Client app (Angular) must provide WebAPI app client id as `resource identifier`. I could not configure it in this library.

I thought of adding it and using it for my own app. If you like it as a configurable setting, you can accept this pull request

Code is fairly basic and self explanatory. Could not find pull request guidelines so putting it all in words.

Cheers and Good luck,
Rahul.


damienbod:
Hi @rahulrulez excellent, thanks for the pull request.

2 things;

the resource option is not part of the OpenID Implict Flow spec. This is just for Azure AD specific. Correct?

http://openid.net/specs/openid-connect-implicit-1_0.html

The "ngc-webpack": "^3.0.0"

Do we need this? @robisim74 

Will merge once the ngc-webpack dependency is clear.  For me looks good.

Greetings and thanks Damien

robisim74:
All the changes in package.json should be removed. Also package-lock.json for now: we should use npm 5 by default.

FabianGosebrink:
thought about introducing yarn: thoughts? Seperate topic?

rahulrulez:
@damienbod 
Hi,

Yes, it indeed looks like Azure AD specific parameter. I did not want to go away from oidc standards. I will make certain changes. Drop down the version of npm and remove unwanted stuff and update here.

Thanks for prompt reply guys :-)

rahulrulez:
I have added a commit that removes unwanted packages from package.json. Also removed the lock file.

damienbod:
@rahulrulez I merge it, and and create a npm package, Might be Wed. before it's goes public. 

Is there any discussion channel for this library? No, I just use github issues. Any suggestions?

@FabianGosebrink yes

rahulrulez:
@damienbod  - Gitter? Github issues is fine for issues. General discussion related to lib can go in a discussion channel.

damienbod:
ok, could try it, I have never given this a go :)

PR #29, opened 2017-06-25
Allow DI of a custom Storage implementation.

astegmaier:
This change to `oidc.security.common` allows it to accept a custom implementation of Storage from angular DI.

For example, you can get angular-auth-oidc-client to store access tokens in Cookies by downloading and adding [Cookie-Storage](https://github.com/bouzuya/cookie-storage) to your project, creating a factory method to provide it:
```
let cookieStorageFactory = () => {
    return new CookieStorage();
}
```
..and then adding it to the providers array in @NgModule:
`{ provide: Storage, useFactory: cookieStorageFactory }`


damienbod:
@astegmaier I'll create a new npm release in the next day or 2 for this. Thanks for the help.

PR #47, opened 2017-07-22
FIX Bug loading default values into configuration when boolean param

openpef:


damienbod:
thanks

damienbod:
Version 1.3.1 

Thanks  a million

PR #53, opened 2017-07-26
Feature/custom state

openpef:


damienbod:
Hi @openpef 

Thanks for this. Why would you want to not send a new state for each Authorize request?

Greetings Damien

openpef:
Hi @damienbod 

In the specifications [https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest), state parameter is recommended for security.

But in many places, when we need to pass some customized informations to the redirect_url, it is suggested to use the state param.
So in this case, we need to read this value and we cannot if it is cleared in the successful_validation method.

That why I added a config param to set if we clear automatically state, if we use it only for security, or not, for the others cases.

Pierre Emmanuel

damienbod:
Hi @openpef

could you rebase this pull request so I can merge? It has conflicts now that I merged your other 2 PRs

Thanks Damien

PR #55, opened 2017-08-02
Fix http headers

HWouters:
The header 'Content-Type', 'application/json' causes the following error:

Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

I encountered this issue while using this library with Azure B2C as authorization server.

### DIFF
--- a/src/services/auth.well-known-endpoints.ts
+++ b/src/services/auth.well-known-endpoints.ts
@@ -97,7 +97,6 @@ export class AuthWellKnownEndpoints {
     private getWellKnownEndpoints = (): Observable<any> => {
 
         let headers = new Headers();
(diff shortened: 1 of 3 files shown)


damienbod:
@HWouters thanks, I'll check and merge this this week. I encountered this problem with google as well, didn't realize I could fix it like this.

Greetings Damien

damienbod:
@HWouters Hi @HWouters tested this in firefox, it does work because a Content-Type text/plain is sent.

HWouters:
@damienbod thanks for merging, works well now.

damienbod:
Thanks as well

PR #57, opened 2017-08-09
Fix authorization url construction

HWouters:
Azure AD B2C authorization endpoint has a query parameter like this:
/oauth2/v2.0/authorize?p=b2c_1_siupin. The url construction couldn't
handle this.
Now I use URL class instead of manual string concatenation.

### DIFF
--- a/src/services/oidc.security.service.ts
+++ b/src/services/oidc.security.service.ts
@@ -399,31 +399,24 @@ export class OidcSecurityService {
         this.oidcSecurityCommon.store(this.oidcSecurityCommon.storage_is_authorized, true);
     }
 
-    private createAuthorizeUrl(nonce: string, state: string): string {
+    private createAuthorizeUrl(nonce: string, state: string) {
 
-        let authorizationUrl = this.authWellKnownEndpoints.authorization_endpoint;
-        let requiredParams = {
-            client_id: this.authConfiguration.client_id,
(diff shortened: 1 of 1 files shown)


damienbod:
Released, version 1.3.4 thanks

damienbod:
Reverted this in 1.3.5 and added test

damienbod:
https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-reference-oidc

https://github.com/damienbod/angular-auth-oidc-client/blob/master/tests/services/oidc.security.service.spec.ts#L89

This works now in version 1.3.5

HWouters:
Hi @damienbod, 
I don't think it will work. For example the following well know configuration url

`https://login.microsoftonline.com/fabrikamb2c.onmicrosoft.com/v2.0/.well-known/openid-configuration?p=b2c_1_sign_in`

returns the following authorization endpoint:

`https://login.microsoftonline.com/fabrikamb2c.onmicrosoft.com/oauth2/v2.0/authorize?p=b2c_1_sign_in`

createAuthorizeUrl appends parameters starting with a ?. This gives a result with 2 questionmarks like this:

`https://login.microsoftonline.com/fabrikamb2c.onmicrosoft.com/oauth2/v2.0/authorize?p=b2c_1_sign_in?client-id=...`

PR #68, opened 2017-08-24
Throws Exception when the library is used in an application inside a iframe (cross domain)

NilsEngelbach:
Implemented check if window.parent is accessible (CORS).

Browsers will throw an Exception if the library is used in an app within an IFrame (from different domain).

Note: Handling of IFrame should be further improved (duplicated code e.g.).


damienbod:
Hi @NilsEngelbach looks, good, I'll merge this, when I get time to tes, should be this weekend

Thanks and Greetings Damien

damienbod:
@NilsEngelbach 

Released 1.3.12

Thanks for the PR!

PR #72, opened 2017-09-01
Add option to trigger event on authorization resolution instead of automatic redirect

CesarD:
Instead of forcing the application consuming this library to automatically redirect to one of the 3 hard-configured routes (start, unauthorized, forbidden), this modification will add an extra configuration option to override such behavior and trigger an event that will allow to subscribe to it and let the application perform other actions.
This would be useful to allow the application to save an initial return url so that the user is redirected to it after a successful login on the STS (ie: saving the return url previously on sessionStorage and then retrieving it during the triggering of the event).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4395 @@
+{
+    "name": "angular-auth-oidc-client",
+    "version": "1.3.12",
+    "lockfileVersion": 1,
+    "dependencies": {
+        "@angular/animations": {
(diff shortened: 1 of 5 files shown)


damienbod:
Thanks for the pull request!, looks good, I test it and merge today.

Greetings Damien

damienbod:
@CesarD released in 1.3.13

Thanks a million!

Greetings Damien

PR #77, opened 2017-09-05
fix: if auto_userinfo is false, we still need to execute runTokenVali…

ydg:
fix: if auto_userinfo is false, we still need to execute runTokenValidatation


damienbod:
@ydg thanks a million!

I merge it and create a new npm today

damienbod:
Version 1.3.15, thanks

Issue #87, opened 2017-10-05
Local Storage session_state undefined parse error


haidelber:
The value of my session_state in local storage is serialized as undefined.

Undefined is not supported in JSON. Therefore JSON.parse throws a SyntaxError as soon as it encounters undefined.

``
ERROR SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at BrowserStorage.read (vendor.js?v=oPJTytp2_H6lNjqzfV8c4mOtYe_1PGijbtehhkuMq44:75071)
    at OidcSecurityCommon.retrieve (vendor.js?v=oPJTytp2_H6lNjqzfV8c4mOtYe_1PGijbtehhkuMq44:75125)
``
![auth_bug](https://user-images.githubusercontent.com/1950208/31231131-4036dd82-a9e6-11e7-8703-5cc6429b3a5d.PNG)

I guess the best way to avoid this issue is to explicitly check for undefined and use null instead. [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description) describes this behavior of JSON.stringify.
I'll submit a pull request in a few minutes fixing this issue.

damienbod:
@haidelber 
fixed and added to version 1.3.18
https://www.npmjs.com/package/angular-auth-oidc-client

thanks for the fix

Gruss Damien

haidelber:
Thanks for generating a new package that fast. Now my angular application works again without any errors.

PR #88, opened 2017-10-05
Serialize undefined as null to storage

haidelber:
Local Storage session_state undefined parse error Issue #87


damienbod:
@haidelber thank you for this fix. I'll test and create a npm as soon as possible.

Greetings Damien

Issue #26, opened 2017-06-23
Reduce size of package


damienbod:
Package depends on jsrsasign which doesn't tree shake so good. See if it's possible to reduce the size of the auth.module

ruimserrano:
Hi,

other thing that don't tree shake very well is rxjs, if you load the bundled file:
import { Observable } from 'rxjs/Rx';

noticed in 
https://github.com/damienbod/angular-auth-oidc-client/blob/40790b8909f5218536c6f4f92d7f8329ac5b225d/src/services/oidc.security.service.ts#L7

on other places, your doing the right thing:
https://github.com/damienbod/angular-auth-oidc-client/blob/40790b8909f5218536c6f4f92d7f8329ac5b225d/src/services/oidc.security.user-service.ts#L3

more info  on why importing rxjs/RX is bad: https://christianliebel.com/2017/07/import-rxjs-correctly/

i could do a PR with this, no prob...glad to contribute

ps: i use https://www.npmjs.com/package/source-map-explorer to check what parts of libs was really bundled. Thats how i spot that i was loading the whole rxjs lib after i added your lib.

damienbod:
@ruimserrano thanks, nice fix

mseknibilel:
Hi,

the jsrsasign is adding about 200 kb to the final bundle (after minification and tree shaking). is there something to do about it please?

elvirdolic:
Some example how it will be solved in this lib: https://github.com/manfredsteyer/angular-oauth2-oidc/pull/356

200kb are really huge, 13% of our bundle

elvirdolic:
@damienbod any idea how do you want to tackle this. I would like to help here but I’m not sure if it will be accepted.

profet23:
as far as I can tell, there are 2 possible approaches:

1. create a custom build/package of jsrsasign.
2. replace jsrsasign with another package.

Can we do a pro/con?

damienbod:
closing this, please open if you still have an issue

elvirdolic:
@damienbod was something done here? I can’t find any commit which solves the problem with jsrsasign . Any special reason to close it?

damienbod:
angular-auth-oidc-client.umd.min.js == 54KB

damienbod:
improved in version 10.0.8. Using a jsrsasign which is reduced in size now.

PR #91, opened 2017-10-10
fix: import from a specific location

maxisam:
for #26

### DIFF
--- a/src/services/oidc.security.service.ts
+++ b/src/services/oidc.security.service.ts
@@ -4,7 +4,7 @@ import { Injectable, EventEmitter, Output } from '@angular/core';
 import { Http, Response, URLSearchParams } from '@angular/http';
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/catch';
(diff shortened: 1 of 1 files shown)


robisim74:
@damienbod They are right that the current import: `import { Observable } from 'rxjs/Rx';`  is wrong.

@maxisam Hi, the pr is not complete, and therefore it is incorrect. It is also necessary to import the individual operators used, such as `timeinterval` etc. in the classes that used them. Then you have to update the `globals `packages in the `rollup.config.js` file.
Please let us know if you are going to complete the request, otherwise I will add the commit to this pr. Thank you!

maxisam:
@robisim74 You can finish it or have a new pr to address this. Normally what I do is like having a separate file for importing all operators I used for rxjs. I don't know which way the maintainer wanna go.

At first, I thought it was just a quick pr, so I just did it without giving too much thought. Feel free to do anything you like to make it work.

damienbod:
@maxisam  @robisim74 thanks, I was offline for the last week, catching up now.

Issue #98, opened 2017-10-16
startup_route -> login_route?


fdcastel:
Suggestion 1: to rename `startup_route` to `login_route` (the one called after a successful _login_).

Suggestion 2: to add a `logout_route` (to be called after a successful _logout_).

Also, the default value `/dataeventrecords/list` is not very useful (except for the sample apps 😉 ). What about to use `'/'` for both login and logout?

damienbod:
Hi @fdcastel te renaming sounds good. Do you want to cretae a pull request? Otherwise I will do it. Thanks. This would be a breakin change, so I will merge it with the release of the HTTP client update, Angular 5 release :)

Suggestion 2 and 3 tip top as well.

Greetings Damien

fdcastel:
Sure. It'll be a pleasure! 😄

damienbod:
Hi @fdcastel I had a few problems improvements with the PR whils testing. 
I have created a branch where we can test and finish the changes. 

https://github.com/damienbod/angular-auth-oidc-client/tree/dev_fdcastel

- I renamed the login_route to post_login_route
- The isAuthorized event is required, because some users of the lib don't use the userData in this way, auto_userData
- I'm not certain the logout is required, we already have a post_logout_uri, need to think about this
- I'm testing this branch now (weekend) and plan the merge then

Thanks fpr the PR.

Greetings Damien

@robisim74 @FabianGosebrink  Any opinions?

damienbod:
Also I will merge the commit from @robisim74, I reset the master to the last released version until this is fixed

robisim74:
Ops... @damienbod I had not noticed the revert, and I re-made the push of the last commits... Please, revert again.

- Ok for `post_login_route`, `logout_route` seems a duplicate
- Ok for keeping isAuthorized

robisim74:
About _post_login_route_, a method would be useful to change it at runtime, and then be able to return to the page requested by the user.

damienbod:
@robisim74 This is possible, the post_login_route is a default route if you're not using the return to page request feature.

robisim74:
`trigger_authorization_result_event`, right.

fdcastel:
> I renamed the login_route to post_login_route
👍 



> I'm not certain the logout is required, we already have a post_logout_uri, need to think about this

My fault. Really `post_logout_redirect_uri` does the job (called from STS server).

[...]

PR #99, opened 2017-10-16
Fix #97 and #98.

### LINKED ISSUE #97: UserData and IsAuthorized out of sync?
After a successful login both `_isAuthorized` and `_userData` observables are changed.

But after a successful logout just `_isAuthorized` is changed. And `_userData` keeps its old value.

This is the expected behavior? If yes, please explain.

If no, then may I suggest to keep just *one* `Observable<UserData>` and simply derive `_isAuthorized` from it? (`isAuthorized => !!userData`).

Finally, I believe [this](https://github.com/damienbod/angular-auth-oidc-client/blob/master/src/services/oidc.security.service.ts#L38) is result from a bad copy & paste and should be removed 😉.

Best regards and thanks for this great piece of code!

fdcastel:
- Renames startup_route to login_route.
- Adds logout_route.
- Do not keep isAuthorized state anymore (mapped from userData).
- Removes OidcSecurityService._userDataValue (unused code).
- Removes OidcSecurityUserService.handleError (unused code).
- Fix minor typos in docs and code.


fdcastel:
It was easier to fix both #97 and #98 together since they are related. Please review and comment.

Apart from the `startup_route` rename, there should be no breaking changes in this PR.

damienbod:
I'll test and merge it as soon as possible, thanks

Issue #79, opened 2017-09-07
More events notifications


CesarD:
I'm dealing with some scenarios where I have my application logged-in, if the access token expires when the application is not running (thus not renewing the token when needed) and when I enter again, the api call will fail because the token is already expired, and there's no way to be notified by the library if this has happened.
Same would apply if there was some error during the token silent renew.

Perhaps you could add some more validation to the token when the library is starting and emit some event or something to better acknowledge of these situations.

damienbod:
https://github.com/damienbod/angular-auth-oidc-client/issues/101

fdcastel:
> if the access token expires when the application is not running (...)

Please, enlighten me: If the app was not running, when you come back and run it will be start in `isAuthorized = false` state. Isn't this enough for you to know that the token needs to be renewed? I can't see a need for another event here.

Now, when the application **is** running, I can see a use for events. In #101 you suggested:

> having some events like "AccessTokenExpiring" or "AccessTokenExpired",

After #101, if silent renew is disabled you will get an `AuthenticationLogout` event. I believe this cover your `AccessTokenExpired` case right?

Now, for `AccessTokenExpiring`, I believe we could emit one [here](https://github.com/damienbod/angular-auth-oidc-client/pull/107/files#diff-3551bf70e40e99dfec0656111f37bf52L573) (and use the already existing `silent_renew_offset_in_seconds` option).

> Same would apply if there was some error during the token silent renew.

This one is also covered: `RefreshTokenError` 😄 (again, after #101).

What you think?

CesarD:
> If the app was not running, when you come back and run it will be start in isAuthorized = false state.

Will it? I think it was returning "true" because the id_token was stored in the browser storage, but that doesn't necessarily reflected that the access_token was still valid. In fact, I don't think there's any place that checks whether the access_token is already expired or not...
A user is "authorized" into the application if his/her token is not expired, otherwise it should be "not authorized" (isAuthorized=false).

> After #101, if silent renew is disabled you will get an AuthenticationLogout event. I believe this cover your AccessTokenExpired case right?

Under what circumstance? 

As for the rest regarding `AccessTokenExpiring` and `RefreshTokenError`, I agree  :)

fdcastel:
> Will it? (...)

Hm. I see. @damienbod ?



> Under what circumstance?

[...]

PR #107, opened 2017-10-22
Adds centralized events property. Fix #101.

### LINKED ISSUE #101: Suggestion: Error handling, events and code cleanup/refactors
I was testing my app against a down STS server and I did notice the error handling for `OidcSecurityService` is lacking. Some considerations:

- `OidcSecurityService.authorize()` doesn't provide any way for the caller to know if something wrong occurred. It has no return values and doesn't throw any exceptions.
  - The only thing it does in case of error is to call `oidcSecurityCommon.logError()`.
- `OidcSecurityService.handleError()` is never called.
  - Due to this, the `authConfiguration.forbidden_route` setting appears to be never used.
- `OidcSecurityService.runGetSigningKeys()` is never called.
  - Due to this, `OidcSecurityService.handleErrorGetSigningKeys()` is also never used.
- In several places we have the following pattern:
  ```javascript
  if (this.authConfiguration.trigger_authorization_result_event) {
      this.onAuthorizationResult.emit(AuthorizationResult.authorized);
  } else {
      this.router.navigate([this.authConfiguration.login_route]);
  }
  ```
  which is somewhat redundant, since the act of `.navigate()` to a certain route could be dispatched from a `onAuthorizationResult` (or similar) subscription.

Please take the above points with a grain of salt, since I'm very new to this code base and maybe I'm missing something. In this case, please enlighten me.

# Suggestions

To minimize breaking changes in the current code I propose the creation of a `.events()` property (very similar to [Angular Router Events](https://angular.io/api/router/Event)).

Suggested events: 

  - `AuthenticationStart`, `AuthenticationSuccess`, `AuthenticationError, AuthenticationRefresh, AuthenticationLogout` (main events)
    - `FetchConfigurationStart`, `FetchConfigurationSuccess`, `FetchConfigurationError` 
      - maps `getWellKnownEndpoints()`
    - `FetchKeysStart`, `FetchKeysSuccess`, `FetchKeysError`
      - maps `getSigningKeys()`
    - `FetchUserInfoStart`, `FetchUserInfoSuccess`, `FetchUserInfoError` 
      - maps `getIdentityUserData()`
    - `RefreshTokenStart`, `RefreshTokenSuccess`, `RefreshTokenError` 
      - maps the timer from `runTokenValidatation()`

This would be a central point for all event-related notifications and the user could subscribe to it to handle whatever he/she wishes. Initially this would be given just as an _additional_ way of reacting to events (i.e. it would not change any of current events/subscriptions). This way it would not require any breaking changes to the existing code (neither introduce any new bugs 😉 ).

Then, with these events in place, we could port independent parts of code (e.g. logging, route navigation, core functionality) from the current state to _subscriptions_ of these events.

IMO this could bring large benefits to the code readability and maintenance.

And, of course, I'm willing to make PRs for all this, if agreed. 😄 

Best regards!

fdcastel:
- Adds events property and typed events.
- Removes _isAuthorizedValue (redundant, BehaviorSubject already stores the last value)
- Removes lastUserData and onUserDataChanged() (not needed anymore)
- Minor refactors and code formatting
- Updates package-lock.json

P.S.: Please review "ToDo:" and "Warning:" comments in the code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "angular-auth-oidc-client",
-    "version": "3.0.0",
+    "version": "3.0.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
@@ -4344,6 +4344,12 @@
                 }
             }
         },
+        "string_decoder": {
+            "version": "0.10.31",
+            "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-0.10.31.tgz",
+            "integrity": "sha1-YuIDvEF2bGwoyfyEMB2rHFMQ+pQ=",
+            "dev": true
+        },
         "string-width": {
             "version": "1.0.2",
             "resolved": "https://registry.npmjs.org/string-width/-/string-width-1.0.2.tgz",
@@ -4354,12 +4360,6 @@
                 "is-fullwidth-code-point": "1.0.0",
                 "strip-ansi": "3.0.1"
             }
-        },
-        "string_decoder": {
-            "version": "0.10.31",
-            "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-0.10.31.tgz",
-            "integrity": "sha1-YuIDvEF2bGwoyfyEMB2rHFMQ+pQ=",
-            "dev": true
         },
         "strip-ansi": {
             "version": "3.0.1",
--- a/public_api.ts
+++ b/public_api.ts
@@ -8,3 +8,4 @@
 
 export * from './src/angular-auth-oidc-client';
 export * from './src/services/authorization-result.enum';
+export * from './src/services/oidc.security.events';
--- a/src/services/auth.well-known-endpoints.ts
+++ b/src/services/auth.well-known-endpoints.ts
@@ -44,7 +44,6 @@ export class AuthWellKnownEndpoints {
 
             if (data.end_session_endpoint) {
                 this.end_session_endpoint = data.end_session_endpoint;
-
             };
 
             if (data.check_session_iframe) {
--- a/src/services/oidc.security.events.ts
+++ b/src/services/oidc.security.events.ts
@@ -1,0 +1,65 @@
+export class AuthenticationStart {
+    constructor(readonly id: number) { }
+
+    toString = () => `AuthenticationStart(id: ${this.id})`
+}
+
+export class AuthenticationSuccess {
+    constructor(readonly id: number) { }
+
+    toString = () => `AuthenticationSuccess(id: ${this.id}')`
+}
+
+export class AuthenticationError {
+    constructor(readonly id: number, readonly err?: any) { }
+
+    toString = () => `AuthenticationError(id: ${this.id}, err: '${this.err}')`
+}
+
+export class AuthenticationLogout {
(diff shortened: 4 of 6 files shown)


damienbod:
@fdcastel thanks a million. Looks good, a lot to test here, as the lib has no auto tests, I do this manaully. I'll plan to complete this at the weekend. 

1 question? Should we document the events in the  API doc?

fdcastel:
> Should we document the events in the API doc?

Surely! I was just waiting for your feedback about the events I implemented (are they good? something missing? something to remove? What about the parameters passed in each one?).

As proposed in #101 I looked to make this PR additive-only, with no breaking changes at all.

damienbod:
I think the events are good. @robisim74 @CesarD

robisim74:
@damienbod @fdcastel I'm not understanding what's going on here: ok for the events, but also this pr has a lot of strange changes.

For example, @fdcastel you are now removing `onUserDataChanged` method, but this method has been added by you 6 days ago with this pr: https://github.com/damienbod/angular-auth-oidc-client/pull/99/files

I think prs should treat one thing at a time, or at least have different commits: you can not test so many changes.

CesarD:
Thanks for tagging me, @damienbod 
Looks good so far (though I'm not having a lot of time to look into the very detail).

My main concern when I created #79 was about having some events like "AccessTokenExpiring" or "AccessTokenExpired", which should trigger when such scenarios are occurring (either during the execution of the app or when loading it). This way, I can hook them from the app to make redirects or showing messages to the users (in the case silent refresh was not possible for any reason). If I cannot detect that the access token expired, I'd be more easily prone to make an unauthorized request to the server without prior notice.

fdcastel:
@robisim74 I fully agree with you.

I'm sorry for the lack of time to polish the PR. I just wrote it in a few hours I had available and send it for feedback. We don't need to merge it as it stands. Also, I did not documented the changes at all (apart from the short commit description, which I assumed was enough).

About the event `onUserDataChanged`: I removed it **exactly** because I just had created it. When Damien brought back `_isAuthorized` (which I deemed unnecessary) from #99 I started to understand better some concepts and changed my mind about some things.

Now I understand how `_isAuthorized` is useful. And now, with the structured events in place, it seems to me `onUserDataChanged` is just not needed anymore.

Please ask about any other strange things you found. I'll be glad to explain it 😉 👍

CesarD:
@fdcastel how about getting the events I mentioned before?

Unless you already accomplish that with the ones you already introduced in #101?

Like I mentioned before, I'd be interested in knowing when the access token in storage is already expired or not so I can take proper action.
Thanks for your work :)

fdcastel:
@CesarD I'm all in to include every event we deem necessary. Let's talk about them in #79.

MovGP0:
removing `_isAuthorized` was next on my list. seems like this has already been done 👍

MovGP0:
```powershell
git remote add upstream https://github.com/damienbod/angular-auth-oidc-client.git
git fetch upstream
git checkout master
git reset upstream/master --hard
git checkout feature/BRANCHNAME
git rebase master
# rebase and fix conflicts
git push --force

fdcastel:
Working on a new PR now from the latest master. Please disregard this one.

Issue #113, opened 2017-10-25
ERROR ReferenceError: 'URLSearchParams' is undefined in IE and Edge


rpipkin:
Guys, I love the direction of this project.  I was using your raw ts files from before it was an npm package.  I recently replace those with this npm package version.  In IE and Edge, issuing a login request creates this bug in the console and never goes to the Identity server login page.  Here's the error with the stack:

ERROR ReferenceError: 'URLSearchParams' is undefined
"ReferenceError: 'URLSearchParams' is undefined
   at OidcSecurityService.prototype.createAuthorizeUrl (http://localhost:5003/dist/app.bundle.js:27407:9)
   at OidcSecurityService.prototype.authorize (http://localhost:5003/dist/app.bundle.js:27127:9)
   at AppComponent.prototype.login (http://localhost:5003/dist/app.bundle.js:82679:9)
   at Anonymous function (Function code:10:11)
   at handleEvent (http://localhost:5003/dist/app.bundle.js:12205:87)
   at callWithDebugContext (http://localhost:5003/dist/app.bundle.js:13666:9)
   at debugHandleEvent (http://localhost:5003/dist/app.bundle.js:13254:5)
   at dispatchEvent (http://localhost:5003/dist/app.bundle.js:8809:5)
   at Anonymous function (http://localhost:5003/dist/app.bundle.js:9420:13)
   at Anonymous function (http://localhost:5003/dist/app.bundle.js:39761:9)"

There is a conversation in Angular's github discussing how URLSearchParams is a global variable and can easily be confused with the Angular class. https://github.com/angular/angular/issues/16378
I have a sneaking suspicion that IE and Edge are finding the wrong object to load.  If I get time I'll try to clone and fix and submit a Pull Request, but if you guys get to it first, it would be hugely appreciated.

robisim74:
@rpipkin thanks for reporting. This is a bug.

@damienbod The error was created by this commit: https://github.com/damienbod/angular-auth-oidc-client/commit/d0745e7a828dd7e319e848acbdeabe83419d998c, when you moved to HttpClient: first `URLSearchParams` was imported from `Http` module, while the local variable is now being used.

robisim74:
Has pr: #114

damienbod:
fixed in 3.0.2, thanks @rpipkin @robisim74

mcano808:
A question- how can I make sure I get this fix into my current angular app?

damienbod:
[...]

PR #114, opened 2017-10-25
Fix URLSearchParams: #113

robisim74:


robisim74:
WIP

robisim74:
@damienbod At the moment, the library doen't work in IE and Edge. When you moved to HttpClient, `URLSearchParams` still worked because there is a JavaScript interface `URLSearchParams`: but it is not supported by all browsers: https://developer.mozilla.org/it/docs/Web/API/URLSearchParams.

I moved to `HttpParams`, but I found another problem (Angular bug): https://github.com/angular/angular/issues/18261, still opened in v5. So as suggested here: https://github.com/angular/angular/issues/18261#issuecomment-338354119, I created a custom encoder with the default `encodeURIComponent` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

It seems works, but it is important that you do other tests.

damienbod:
@robisim74 thanks, I'll merge this first then with the doc PR.

damienbod:
tested:

Chrome:

t4=ABC%20abc%20123&t3=%23&t2=-_.!~*()&t1=;,%2F%3F%3A@%26%3D%2B$

Edge:

t4=ABC%20abc%20123&t3=%23&t2=-_.!~*()&t1=;,%2F%3F%3A@%26%3D%2B$

IE:

t4=ABC%20abc%20123&t3=%23&t2=-_.!~*()&t1=;,%2F%3F%3A@%26%3D%2B$

Firefox:

t4=ABC%20abc%20123&t3=%23&t2=-_.!~*()&t1=;,%2F%3F%3A@%26%3D%2B$

thanks, also added a test

Issue #14, opened 2017-06-17
support prompt request parameter


damienbod:
Request with prompt=login

Request with prompt=none when not logged in

Request with prompt=none when logged in

http://openid.net/wordpress-content/uploads/2016/12/OpenID-Connect-Conformance-Profiles.pdf

damienbod:
added

PR #122, opened 2017-10-31
Silent renew: prompt=none

robisim74:


robisim74:
@damienbod This pr:
- Add `prompt= none` to silent renew, according to the spec: in fact some op do not refresh the token in the absence of it. Related to: https://github.com/damienbod/angular-auth-oidc-client/issues/14
- Fix the starting of silent renew and check session after the _authWellKnownEndpoint_ has been loaded, to avoid an undefined router (they use its info)
- Fix(building): public api exports

damienbod:
tested, thanks

PR #135, opened 2017-12-04
Update oidc.security.service.ts

eddiederk:
Allow developers to manually call refreshSession() to reload the users Token


damienbod:
Will create a release tonight or tomorrow

PR #144, opened 2018-01-01
UserData should be any instead of string

haidelber:
I just upgraded to the newest version an realized the top level API changed a bit. The getUserData() should be any in my opinion instead of string. At least in my project it's definitely an object in the form of {sub: "", email: "", roles: Array(2)}.

### DIFF
--- a/src/services/oidc.security.service.ts
+++ b/src/services/oidc.security.service.ts
@@ -126,7 +126,7 @@ export class OidcSecurityService {
         }
     }
 
-    getUserData(): Observable<string> {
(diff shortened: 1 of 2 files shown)


damienbod:
Yes, thanks for this catch. Will release this today as soon as the build finishes.

damienbod:
Thanks will be released in version 3.0.11

damienbod:
fixed

haidelber:
Thanks as always for releasing super fast :)

PR #168, opened 2018-02-01
Fix a couple of issues relating to silentRenew

runninghare:
1. Replace the timer observable with a recursive setTimeout in order to prevent firing another event when the previous tokenRenew process has not yet finished.
2. Encapsulate the silentRenew heartbeat check into ngZone.runOutsideAngular to avoid triggering change detection.

### DIFF
--- a/src/services/oidc.security.service.ts
+++ b/src/services/oidc.security.service.ts
@@ -5,8 +5,7 @@ import { EventEmitter, Injectable, Output } from '@angular/core';
 import { Router } from '@angular/router';
 import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import { Observable } from 'rxjs/Observable';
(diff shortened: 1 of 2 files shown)


damienbod:
cool, I'll test this tomorrow, thanks!

damienbod:
@runninghare thanks, the code looks good, testing it now.

damienbod:
all good thanks

damienbod:
@runninghare 
Also need to fix the same problem in the OidcSecurityCheckSession. Would you like to do a PR for this as well? If not, I'll update it. Thanks for these fixes. I'll do a release this weekend

runninghare:
I just created another pull request for OidcSecurityCheckSession. However I haven't used this before so please check whether the update makes sense.

Also did a little refactoring on OidcSecurityService.

PR #191, opened 2018-03-05
allow custom URL handlers in authorize()

nonAlgebraic:
In `OidcSecurityService`, the `authorize()` method builds an authorization URL and then immediately sets the `window`'s `location.href` to that URL, navigating the user away from the Angular app context onto the OpenID Provider's authorization endpoint. This commit modifies that method so that it accepts an optional handler function as an argument. If a handler is provided, the authorization URL will be passed to it instead of triggering a window navigation.

Use case:
An implementing app wants to initiate the authorization process in a popup or a silent iframe. Currently, the only way to achieve that is to load a page in the popup/iframe which would call the `authorize()` method internally, which requires loading and bootstrapping a whole additional instance of Angular. Instead, the app should be able to load the authorization URL directly in the popup/iframe.

### DIFF
--- a/src/services/oidc.security.service.ts
+++ b/src/services/oidc.security.service.ts
@@ -180,7 +180,7 @@ export class OidcSecurityService {
         this.oidcSecurityCommon.customRequestParams = params;
     }
 
-    authorize() {
+    authorize(urlHandler?: (url: string) => any) {
         if (this.authWellKnownEndpoints) {
             this.authWellKnownEndpointsLoaded = true;
         }
@@ -223,7 +223,11 @@ export class OidcSecurityService {
             state,
             this.authWellKnownEndpoints.authorization_endpoint
         );
-        window.location.href = url;
+        if (urlHandler) {
+            urlHandler(url);
+        } else {
+            window.location.href = url;
+        }
     }
 
     authorizedCallback(hash?: string) {


damienbod:
@nonAlgebraic super thanks!. I'll test it and release it as soon as possible

damienbod:
will be released in version 4.0.3, would you like to add some documentation as well? Thanks for the PR

Greetings Damien

nonAlgebraic:
here it is. This my first two pull requests ever, hopefully I've gone about it properly! #192

satkanth:
Is there an example of how this is done?

nonAlgebraic:
@satkanth I'm guessing that your app is loading in the new window because whatever route component is loaded at the redirect URL is still set up to behave as if it were the only app instance, loaded in the parent window.

There are several ways to implement the popup/iframe method. As an example, you will need to set up:

- A component that calls `authorize()` (let's call this one `LoginComponent`)
- A component that loads at the callback URL (let's call it `CallbackComponent`)

The basic process would go as follows (you will need to flesh this out before using):

1. `LoginComponent` will call `authorize()` with a handler function that loads the Provider's authentication URL in a popup:
```TypeScript
this.oidcSecurityService.authorize((authUrl) => {
        // handle the authorrization URL
        const popup = window.open(authUrl, '_blank', 'toolbar=0,location=0,menubar=0');
    });
```

2. `CallbackComponent` will be loaded in the popup once the Provider redirects back to it and - if the authentication was successful - will have access to the provided OAuth tokens. It will then send those tokens back to the parent window using `postMessage()`:
```TypeScript
if (window.location.hash) {
  window.opener.postMessage(
    window.location.hash.substr(1), // token from the URL without the `#` symbol
    targetOrigin                    // "*" or the app's URL (more secure)
  );
}
```

3. Back in the parent window,  'LoginComponent' will be listening for a "Message" event. When it receives one from `CallbackComponent` in the popup, it will pass it to `OidcSecurityService` for processing:
```TypeScript
window.addEventListener('message', (event: any) => {
  popup.close();
  
  /**
   * authorizedCallback() takes an optional argument `hash` which,
   * if passed, is used as the OAuth token instead of being
   * taken from the current URL.
   */
  this.oidcSecurityService.authorizedCallback(event.data);
});
```

Hopefully this is helpful for you!

satkanth:
Thanks @nonAlgebraic for that detailed answer, really helpful.

Issue #102, opened 2017-10-18
Version 1.3.19: Duplicate of application in iframe for silent renew


mwymann:
Hi, 

I just installed version 1.3.19 and realized in our app, that after the first silent renew the iframe with id="myiFrameForSilentRenew" contains a duplicate of the entire application and initializes it.

Not sure if this is the expected behavior.

Can you please check that?

jcannata:
I am seeing this too on my angular app as well as the AngularClient in damienbod/AspNet5IdentityServerAngularImplicitFlow.
I'm a newbie dealing with OAuth/OIDC so this library has been a real lifesaver!

damienbod:
https://github.com/damienbod/angular-auth-oidc-client/issues/128

damienbod:
@mwymann @jcannata At present the whole app is loaded in the iframe. This coould be improved

jcannata:
@damienbod For the record, I am using version 3.0.4 sources. A related issue to having the entire app loaded in the iframe is that subscriptions such as

```
 this.oidcSecurityService.getIsAuthorized().subscribe(authorized => {
            this.isAuthorized = authorized;
        });
```

in the 'main' app will not receive any change notifications detected by the iframe's instance.
In my case I protect various routes by subscribing to getIsAuthorized(). If a user logs outs (via oidcSecurityService.logoff()) I will receive a notification and everything works correctly. However, if I revoke access to the client directly from the Identity Server the iframe instance will detect this change and update the sessionStorage data but the 'main' instance will still allow access as getIsAuthorized() never fired.

Maybe I'm not understanding how best to use the oidcSecurityService subscriptions? Although not optimal, is there a way to force recalculating the isAuthorized state to account for this situation?

robisim74:
@damienbod I think I find the reason why the whole app is loaded in the iframe. After the silent renew, you redirect to the base `redirect_url`, and so all the content at the redirect url is loaded into the iframe. Other libraries like _oidc-client-js_ redirect to another content, for example _silent-renew.html_, just to provide the necessary operations, so only the content of _silent-renew.html_ is loaded into the iframe.

jcannata:
[...]

PR #199, opened 2018-03-19
Lightweight silent renew

jcannata:
Addresses issue #102 by registering a DOM event which can receive and process an id_token without reloading the entire angular app.


shanehall:
Looking forward to this. Just wondering if it possible to simplify the silent renew file? I have done what you're proposing manually and the html file simply does;

<script>
    parent.postMessage(location.hash, location.origin);
</script>

The handling in the app is similar to what you're doing now.

jcannata:
I am no expert in DOM but I chose dispatchEvent because it directs a message to a specific event (as opposed to the generic 'message' event handler), allowing for other event handlers in the future. Also, I come from the strictly typed language world ;-)
I'm open to changing it. Any thoughts, @damienbod ?

damienbod:
works great, thanks

damienbod:
This is a major feture, will make this as 4.1.0 release

damienbod:
@jcannata Big big thanks! Everything likes good now, just want to do some final testing before I release.

Greetings Damien

Issue #235, opened 2018-05-31
silent-renew problems when using webpack hot module replacement


profet23:
I'm having an issue when webpack runs its hot module replacement.

What's happening is that when the replacement happens, the oidc library has a second instance loaded. When that happens there are now two listeners for the `CustomEvent` `oidc-silent-renew-message`. This causes a race when the silent renew happens and then a failure to renew. 

I believe the library should be adding an instance id to the state and only processing the `CustomEvent` if the instance id matches. If the instance id does not match, it should unsubscribe.

profet23:
PR submitted here: #236

damienbod:
@profet23 thanks

damienbod:
release 6.0.1

PR #236, opened 2018-06-01
Fix for bug #235: silent-renew problems when using webpack hot module…

profet23:
… replacement 

This fix deregisters other listeners of `oidc-silent-renew-message` any time a new one is added. It doesn't make sense to respond to this event from multiple instances.


damienbod:
@profet23 I'll do some tests and do a new release, hopefully at the weekend, I want to add, clean up some open issues with the docs as well.

keyrox:
Awesome!

PR #246, opened 2018-06-28
Throw error when userinfo_endpoint is not defined (Azure AD)

patrickgardella:
Log error, and recommend setting auto_userinfo to false in config.

Ideally, the error would be trapped, and the proper `AuthorizationResult` would be returned to the user. This should be enough to notify the user of the problem, and point them to how to solve it.

### DIFF
--- a/src/services/oidc.security.service.ts
+++ b/src/services/oidc.security.service.ts
@@ -311,6 +311,11 @@ export class OidcSecurityService {
                                 ]);
                             }
                         }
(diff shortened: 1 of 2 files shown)


damienbod:
thanks, nice, I'll create a new npm and test as soon as possible.

damienbod:
release 6.0.2

damienbod:
released
thanks

Issue #297, opened 2018-09-29
silent_renew inconsistent with execution


profet23:
The silent_renew process is behaving inconsistently in my application. Sometimes it will run fine, other times it does not renew.

I have managed to track down the issue to the `silentRenewHeartBeatCheck`. It seems like `this._userData.value` is sometimes blank and because of this, silent renew will not take place. If I go ahead and subscribe to the `getUserinfo()` function that will populate the `_userData.value` and silent renew will function properly from that point on.

If it helps, I'm using the `auto_userinfo` as true in my configuration.

Again, this is only happening sometimes, but often enough. I can't get it to behave consistently. For now I'm just adding a subscription to  `getIsAuthorized` and calling `getUserInfo` there, but this feels hacky. 

Any tips would be appreciated.

profet23:
More information:

It seems that if the initial token retrieved from storage is already expired, the "runTokenValidation()" function is never called. This function kicks off the silent-renew heartbeatcheck. So even if the initial silent renew works and grabs a valid token from the authority, the heartbeat is never actually kicked off. 

Once again, if you manually subscribe to OidcSecurityService.getUserinfo() at any point, it'll begin the heart beat and everything will work as normal.

PR #298, opened 2018-09-29
Removing unnecessary gating of `runTokenValidation()`.

profet23:
`runTokenValidation()` has it own internal gating. The unnecessary gating can lead to situations where silent_renew heartbeat is never started.

Fix for #297


damienbod:
Thanks @profet23 

doing some manaully testing, and will try to merge the PR from @FabianGosebrink before the next release

damienbod:
this solves https://github.com/damienbod/angular-auth-oidc-client/issues/297, right?

profet23:
Correct

Issue #299, opened 2018-10-03
OidcSecurityService.getIsAuthorized() return false after Hot Module Reload


xalhera:
Since version 6.0.9, when webpack updates modified modules, the value returned by OidcSecurityService.getIsAuthorized() is false while the token is valid.

_Before 6.0.9 everything worked well._

The problem is located during AuthGuard.canActivate() method :

```
public canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> | boolean {
        return this.oidcSecurityService.getIsAuthorized()
            .pipe(
                switchMap((isAuthorized: boolean) => {
                    console.log('From [AuthGuard.canActivate] (getIsAuthorized)', isAuthorized);
                    if (isAuthorized) {
                        ...
                    }
                    ...
                    this.router.navigate(['/auth/signin']);
                    ...
                })
            );
    }
```
Because isAuthorized is always false the user is redirected to /auth/signin page.
Here are the debug logs :

```
From [AuthGuard.canActivate] (getIsAuthorized) false
Information: Connection disconnected.
onUserDataChanged: last = undefined, new = 
onUserDataChanged: last = , new = [object Object]
IsAuthorized setup module
[...]

Issue #291, opened 2018-09-20
Cannot authenticate version 6.0.10


xaviergxf:
Hi Damien,

  After upgrading the angular-auth-oidc-client from 6.0.6 to 6.0.10 the authentication is not working anymore. My configuration is as follows:

```
export class AppModule {
    constructor(
        private oidcSecurityService: OidcSecurityService,
        private oidcConfigService: OidcConfigService,
    ) {
        this.oidcConfigService.onConfigurationLoaded.subscribe(() => {

            const openIDImplicitFlowConfiguration = new OpenIDImplicitFlowConfiguration();
            openIDImplicitFlowConfiguration.stsServer = environment.securityTokenServer;
            openIDImplicitFlowConfiguration.redirect_url = `${window.location.origin}`;
            // The Client MUST validate that the aud (audience) Claim contains its client_id value registered at the Issuer
            // identified by the iss (issuer) Claim as an audience.
            // The ID Token MUST be rejected if the ID Token does not list the Client as a valid audience,
            // or if it contains additional audiences not trusted by the Client.
            openIDImplicitFlowConfiguration.client_id = 'Appid';
            openIDImplicitFlowConfiguration.response_type = 'id_token token';
            openIDImplicitFlowConfiguration.scope = 'openid profile email';
            openIDImplicitFlowConfiguration.post_logout_redirect_uri = `${window.location.origin}`;
            openIDImplicitFlowConfiguration.start_checksession = true;
            openIDImplicitFlowConfiguration.silent_renew = true;
            openIDImplicitFlowConfiguration.silent_renew_url = `${window.location.origin}/silent.html`;
            openIDImplicitFlowConfiguration.post_login_route = '/monitor';
            // HTTP 403
            openIDImplicitFlowConfiguration.forbidden_route = '/forbidden';
            // HTTP 401
            openIDImplicitFlowConfiguration.unauthorized_route = '/home';
            openIDImplicitFlowConfiguration.trigger_authorization_result_event = true;
            openIDImplicitFlowConfiguration.log_console_warning_active = true;
            openIDImplicitFlowConfiguration.log_console_debug_active = false;
            // id_token C8: The iat Claim can be used to reject tokens that were issued too far away from the current time,
            // limiting the amount of time that nonces need to be stored to prevent attacks.The acceptable range is Client specific.
            openIDImplicitFlowConfiguration.max_id_token_iat_offset_allowed_in_seconds = 20;

[...]

Issue #257, opened 2018-08-01
Event when everything loaded including tokens if applicable?


JohnGalt1717:
Hi there.

I'm attempting to set up our system so that if the user is already logged in they redirect to their home screen. But only if they are already logged in on the first visit.

This should be trivial but is deceptively difficult.

The onModuleSetup method calls before everything is loaded so you can't subscribe to that.

If you subscribe to the rest of the events there is no consistency to when they fire.

i.e. they will fire every time the site is reloaded even on login completions. (and getIsAuthorized() as an example fires twice on startup, once with false and once with true if logged in, so there is no way to know if the site is done loading completely because if they're not logged in you only get the false)

I'm trying to trap a way that ignores login completions because obviously that should be handled separately and just handle the user coming back to the site with a fresh session and the login info is there.

Suggestions?

profet23:
I use this in order to check authorization:

```
this.isAuthorized$ = new Observable<boolean>((subscriber) => {
            let subscription = this.oidcSecurityService.onModuleSetup.pipe(take(1)).subscribe(() => {
                    subscriber.next(true);
                    subscriber.complete();
                });

            if (this.oidcSecurityService.moduleSetup) {
                subscriber.next(true);
                subscriber.complete();
                subscription.unsubscribe();
            }
        }).pipe(
            switchMap(() => {
                let race$ = race<any>(
                    this.oidcSecurityService.getIsAuthorized().pipe(filter((isAuthorized) => isAuthorized == true)), 
                    this.oidcSecurityService.onAuthorizationResult.asObservable()     
                );

                this.oidcSecurityService.refreshSession();
                
                return race$;
            }),
            switchMapTo(this.oidcSecurityService.getIsAuthorized()),
            shareReplay(1)
        );
```

@damienbod  I feel like this comes up a lot. maybe something like the above needs to be incorporated?

JohnGalt1717:
@profet23 @damienbod 100% . This needs to be a standard event that is fired after all of this. And there needs to be an option to refresh an expired token before this is fired so that it will go and try and do a silent redirect through (if the login site uses cookies and skips through) and then and only then fires this, and can automatically redirect to login if unable to refresh it that way.

What you want is a site to ONLY show up in the logged in or not state. This event should fire passing false if not logged in after all of this stuff too. We would then use this to kick off the UI display and loads of data from there according to logged in or not. 

[...]

PR #302, opened 2018-10-04
Refactoring getIsAuthorized()

profet23:
Refactoring `getIsAuthorized()` to wait until setup and refresh is complete before emitting.

### Previously:
`getIsAuthorized()` would emit its default value of `false` even before setup of the module had been completed. This lead to some very strange race conditions and numerous bug reports.

### With this change:
`getIsAuthorized()` will now:

1.  Wait for setup to complete
2.  Then:
     1. If silent renew _is not_ enabled:
          1. Return the result of authenticating any local tokens either successfully or unsuccessfully. 
     2. If silent renew _is_ enabled then SIMULTANEOUSLY:
          1. Authenticate locally stored tokens and return the result _ONLY_ if authentication is successful. _This action will never return if the locally stored tokens aren't successfully authenticated._
          2. Try and authenticate with the authority and return true or false based on outcome.
3. As soon as one of the above actions complete, the observable will emit and you'll have the authentication result.
4. This result is replayed to any late subscribers.

**While I think the impact will be minimal, this does create a fundamental change to an external facing API so care should be taken when upgrading.**

This issue has come up multiple times:
#257 
#291
#299

### DIFF
--- a/src/services/oidc.security.service.ts
+++ b/src/services/oidc.security.service.ts
@@ -3,7 +3,7 @@ import { HttpParams } from '@angular/common/http';
 import { EventEmitter, Inject, Injectable, NgZone, Output, PLATFORM_ID } from '@angular/core';
 import { Router } from '@angular/router';
 import { BehaviorSubject, Observable, throwError as observableThrowError } from 'rxjs';
-import { catchError } from 'rxjs/operators';
+import { catchError, filter, shareReplay, switchMap, switchMapTo, take, tap, race } from 'rxjs/operators';
 import { OidcDataService } from '../data-services/oidc-data.service';
 import { AuthWellKnownEndpoints } from '../models/auth.well-known-endpoints';
 import { AuthorizationResult } from '../models/authorization-result.enum';
@@ -31,9 +31,12 @@ export class OidcSecurityService {
 
     checkSessionChanged = false;
     moduleSetup = false;
+
+    private _isModuleSetup = new BehaviorSubject<boolean>(false);
+
     private authWellKnownEndpoints: AuthWellKnownEndpoints | undefined;
     private _isAuthorized = new BehaviorSubject<boolean>(false);
-    private _isAuthorizedValue = false;
+    private _isSetupAndAuthorized: Observable<boolean>;
 
     private lastUserData: any;
     private _userData = new BehaviorSubject<any>('');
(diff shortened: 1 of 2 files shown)


damienbod:
@profet23 Thanks for this. After testing, I have problems with this change. I moved these changes to the Refactoring_getIsAuthorized branch for further testing.

I'm testing now to find out the problem.  When the silent renew is active, the oidc loops and produces incorrect warnings. I will test later to find out exactly why. 

If you want to see my test repo which fails:

https://github.com/damienbod/AspNet5IdentityServerAngularImplicitFlow/tree/dev

in the dev branch.

Greetings Damien

damienbod:
@profet23 Looks like your changes are not the problem. Sorry with the latest version of the STS, the error: "consent_required" is now required per default, ans so the silent renew fails, because it is not possible to do a silent renew.

Will test and fix, but your changes look good to go.

Thanks and sorry for the last comment

damienbod:
Have a different problem now :)
error: "login_required
The silent renew should not run until the user has logged in.

damienbod:
Problem is that the this.refreshSession(); is getting called, even though the user has never logged in.

damienbod:
@profet23 When I remove the  this.refreshSession(); from your code, everything works again. Can I remove this, or what was the reason you added it?

Greetings Damien

profet23:
I get the error "login_required' as well. but that was intentional as the `getIsAuthorized()` simply returns false in that situation. And then I thought it would be up to the developer to kick off the authentication.

As an example I have this in a guard. The guard returns getIsAuthorized() and taps it. If the observable emits false then I navigate the user to the signin page of my app which kicks off the login process.

I'm using the silent renew to authenticate a user silently who already has logged into the auth server. For example I have my app running on multiple domains. If the user logs into the first domain, when they open another domain the app will retrieve credentials via silent renew. 

TLDR:  You'll get the error "login_required" but the observable should emit false in that situation and allow the developer to react.

damienbod:
OK, I get a loop and the user cannot react. If I removed the this.refreshSession();, would it still work for you? This fixes my test env. 

Also the message 'OidcSecurityCheckSession pollServerSession session_state is blank' should be a debug log because this is a normal possibility. I will change it the logDebug, good?

And also thanks a million for you help.

Greetings Damien

profet23:
@damienbod the `this.refreshSession()` is what kicks off emitting of the `onAuthorizationResult` event in the second part of the race.

It allows the app to silently authenticate a user who is already authenticated with the server. If they are not authenticated with the server, then the observable should (and I believe does) emit false and the developer needs to call authorize().

Without the `this.refreshSession()` you'll never be able to silently (via iframe) authenticate a user who is already authenticated. 

Maybe this isn't the correct workflow? Not sure. 

As for your loop, my guess is that something else in your app is subscribing either to the `onAuthorizationResult` event ?

profet23:
But the refreshsession() will run only after 10+ seconds. We're looking to get the session refreshed right away as getIsAuthorized() will be used in an guards.

damienbod:
@profet23 yes you are correct.  Reason I was looping, was because I was redirecting if not authorized, and the refresh function was called again. Navigated to the unauthorized, and not a reload, and then the solution worked, but I do 2 login requests with everything initial login, and 1 fails on the callback, the second is ok, because we use the correct nonce then. (1 request for the login, 1 for the refresh session call) Do you have these warnings, problem as well?

profet23:
Yeah I think the refreshSession() is required because without it, getIsAuthorized() may never emit or at least may take an extended amount of time to emit.

While the error message isn't ideal, it is correct. A login is required, so the observable emits false and the developer can then react by calling authorize() if appropriate.

With this change (and adding back the refreshSession() call) a lot of logic should become easier for the developer in that they are mostly just listening for getIsAuthorized() emissions and reacting.

As for the checkForExistingTokensOnInitialOpen I have it doing that by checking if silent renew is configured. I figured if the user has silent renew configured they'd want to silently authenticate.

damienbod:
@profet23 I will try to add it back. Just need to deactivate the silent renew refreshsession call, when the login authorize is running and then everything is fine.

PR #305, opened 2018-10-06
Allow multiple auth states and auth nonces to exist.

profet23:
This is a fundamental change to how auth staes and nonces work.

Multiple auth states and nonces will now be stored. This will allow multiple auth processes to work simultaneously (multiple tabs, auth and refresh, etc.).

Auth states and nonces are valid for 1 hour. Nonces are valid only for one retrieval from storage and are immediately deleted.

### DIFF
--- a/src/services/oidc-security-state-validation.service.ts
+++ b/src/services/oidc-security-state-validation.service.ts
@@ -27,8 +27,7 @@ export class StateValidationService {
         const toReturn = new ValidateStateResult('', '', false, {});
         if (
             !this.oidcSecurityValidation.validateStateFromHashCallback(
-                result.state,
-                this.oidcSecurityCommon.authStateControl
+                result.state
             )
         ) {
             this.loggerService.logWarning('authorizedCallback incorrect state');
@@ -59,8 +58,7 @@ export class StateValidationService {
 
         if (
             !this.oidcSecurityValidation.validate_id_token_nonce(
-                toReturn.decoded_id_token,
-                this.oidcSecurityCommon.authNonce
+                toReturn.decoded_id_token
             )
         ) {
             this.loggerService.logWarning('authorizedCallback incorrect nonce');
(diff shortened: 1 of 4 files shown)


damienbod:
@profet23 thanks. This is a big change and looks good. I will merge and do the testing then.

damienbod:
@profet23 After testing this, I have more problems than before. The silent renew can no longer be turned off, using invalid nonces no longer fails, and the login request still has the same problem. 

I have created a new branch with your changes to continue testing.

I will add the refeshsession back but I will fix the problem at the source of the problem. 

1. this should only run once if not logged in
2. Should never run, if a login request has been started
3. Should not start the silent refresh

I will refactor the refreshsession function and not the nonce states.

Makes sense?

Greetings Damien

damienbod:
@profet23 I have added a fix now for this in the master. Can you test if this works for you?

I test the other flows now, and if all is ok, will make a package. WIll wait for your feedback.


Thanks for all the PRs. This solves a lot of problems.

Greetings Damien

profet23:
@damienbod i think this feature should probably go into my own app's custom storage mechanism. Let's abandon it for now.

profet23:
@damienbod I looked over your changes and submitted my proposed changes here: #306 Let's continue discussion there.

damienbod:
@profet23 Ok, I have merged your changes, will do some more tests. Also, I have 1 last thing to fix, before we can merge this, the user profile needs to be loaded, if the session refresh is called from an existing token. Then it's ready to go.

damienbod:
@profet23 Everything is working now, did all the tests, and have no problems. I'm ready for a release

Good for you?

Greetings Damien

profet23:
Yep! Thanks for being so responsive/patient with my proposals. I think we can also close a lot of the issues that are relevant to this change. 

The documentation for `getIsAuthorized()` probably needs to be updated.

damienbod:
Looking to see what we should change in the docs.

Do we need to change this, or is this guard example still correct?

https://github.com/damienbod/angular-auth-oidc-client#using-guards

No problem, thanks for all the changes, really grateful for your help.

Also, what would you recommend for changes here:

https://github.com/damienbod/angular-auth-oidc-client/blob/master/API_DOCUMENTATION.md#getisauthorized-observable

Thems seems ok to me.

Greetings Damien

PR #307, opened 2018-10-07
Readme and API_Documentation update

profet23:
`take(1)` isn't necessary for guards.

Adding note about `getIsAuthorized()`

Adding documentation for `getIsModuleSetup()`

Adding note about `getIsModuleSetup()`


profet23:
@damienbod not sure if the other setup examples should be redone to use the new `getIsModuleSetup()` method, but I added notes about it.

I do believe the new method is simpler to implement with less code.

PR #314, opened 2018-10-11
If the authority server goes down, the oidc.security.check-session se…

### LINKED ISSUE #310: Is there away to show user when identity server is down
Is it possible to notify the user when identity server is down after they successfully logged in. Do we have to wait till the silent renew or is there a heart beat functionality.

profet23:
…rvice will now emit an onCheckSessionChanged event.

The server will be checked every sixty seconds.
If a message isn't received from the check session iframe within 3 "beats" of the check session heart beat. An event will be emitted.

The iframe will only stop emitting events after the server is checked. So the event will be emitted at a minimum of 3 beats (9 seconds) and  maximum of sixty seconds + 3 beats (69 seconds).

Fixes #310

### DIFF
--- a/src/services/oidc.security.check-session.ts
+++ b/src/services/oidc.security.check-session.ts
@@ -1,5 +1,6 @@
 import { EventEmitter, Injectable, NgZone, Output } from '@angular/core';
-import { Observable, Observer } from 'rxjs';
+import { Observable, Observer, from } from 'rxjs';
+import { take } from 'rxjs/operators';
 import { AuthWellKnownEndpoints } from '../models/auth.well-known-endpoints';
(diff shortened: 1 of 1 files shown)


damienbod:
@profet23 fix works great, thanks!

WIll release a new version today, version 7.0.0. @FabianGosebrink changes in the events is a breaking change, we still need some docs before I do the release.

Thanks

tpatikirikorala:
hi,

Is this fix available, in which version of lib?

profet23:
It should be 7.0.0 @tpatikirikorala

damienbod:
Yes, this has been released in version 7.0.0 yesterday.

PR #317, opened 2018-10-17
Check session will now be stopped when the user is logged out and res…

### LINKED ISSUE #316: QUESTION: Getting OidcSecurityCheckSession not receiving check session response messages. Server unreachable?
I'm trying to use start_checksession so that I can redirect the user to the home page when they logoff from a different browser tab or window, but I keep getting this error even when the user is not logged in: 
OidcSecurityCheckSession not receiving check session response messages. Server unreachable?

I'm not too sure if I'm using the start_checksession correctly but should it still be checking the session even if you're not logged in?

I'm using version 7.0.0

profet23:
…umed when the user is logged in.

Fixes #316


damienbod:
thanks, will test and do a release today

PR #325, opened 2018-10-23
Unicode special characters (accents and such) in JWT are now properly…

### LINKED ISSUE #324: GetPayLoadFomIdToken special characters
I've been having issues with getting the payload from the token, which in some cases can contain special characters like é or ã for names. The problem is that the conversion that's being done reads characters form a different characterset, as far as I can see.

The problem seems to be in src/services/oidc-token-helper.service.ts , line 74
 `return typeof window !== 'undefined' ? window.atob(output) : new Buffer(output, 'base64').toString('binary');`

When I take one of your unittests (tests/oidc-token-helper.service.spec.ts, line 108): 

`it('returns payload if token is correct, encode is false', () => {
	const token =
'SGVsbG8gV29ybGQgMTIzIQ==.eyAidGV4dCIgOiAiSGVsbG8gV29ybGQgMTIzISJ9.SGVsbG8gV29ybGQgMTIzIQ==';
	const expected = JSON.parse(`{ "text" : "Hello World 123!"}`);
	const result = tokenHelperService.getPayloadFromToken(token, false);
	expect(expected).toEqual(result);
});`

and I replace the token with the following, (created @ jwt.io): 
`eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiSm9obiBEw7PDqyJ9.wMn-1oLWnxKJolMGb7YKnlwjqusWf4xnnjABgFaDkI4`

as well as the expected result with 
`{
  "name": "John Dóë"
}`

the test will fail, as the payload result is not the same.

Is it something wrong on my end or an actual bug?

profet23:
… decoded.

Added test.

Fixes #324

### DIFF
--- a/src/services/oidc-token-helper.service.ts
+++ b/src/services/oidc-token-helper.service.ts
@@ -67,7 +67,14 @@ export class TokenHelperService {
                 throw Error('Illegal base64url string!');
         }
 
-        return typeof window !== 'undefined' ? window.atob(output) : new Buffer(output, 'base64').toString('binary');
+        let decoded = typeof window !== 'undefined' ? window.atob(output) : new Buffer(output, 'base64').toString('binary')
+
+        // Going backwards: from bytestream, to percent-encoding, to original string.
+        decoded = decodeURIComponent(decoded.split('').map(function(c) {
+            return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
+        }).join(''));
+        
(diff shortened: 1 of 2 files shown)


damienbod:
@profet23 Thank you very much for all your help. You work has improved this lib massively!

I'll test and do a package.

Greetings Damien

PR #327, opened 2018-10-24
replaced eventemitters with Subjects/Observables and updated and docs

FabianGosebrink:


src/services/oidc.security.service.ts
```diff
@@ -24,12 +24,9 @@ import { UriEncoder } from './uri-encoder';
 
 @Injectable()
 export class OidcSecurityService {
-    @Output()
-    onModuleSetup = new EventEmitter<boolean>();
-    @Output()
-    onAuthorizationResult = new EventEmitter<AuthorizationResult>();
-    @Output()
-    onCheckSessionChanged = new EventEmitter<boolean>();
+    onModuleSetup = new Subject<boolean>();
```
profet23: Subjects should be private

src/services/oidc.security.config.service.ts
```diff
@@ -1,12 +1,11 @@
...
-import { of } from 'rxjs';
+import { Injectable } from '@angular/core';
+import { of, Subject } from 'rxjs';
 import { catchError, map } from 'rxjs/operators';
 
 @Injectable()
 export class OidcConfigService {
-    @Output()
-    onConfigurationLoaded = new EventEmitter<boolean>();
+    onConfigurationLoaded = new Subject<boolean>();
```
profet23: You don't want public Subjects.

FabianGosebrink:
Hey profet23, thank you for your responses and comments. Really appreciate it. I wanted to get rid of the `@Output()` and the `EventEmitter`s as they are closely related to templates and components. However, why Subjects should not be public?

src/services/oidc.security.check-session.ts
```diff
@@ -21,9 +21,7 @@ export class OidcSecurityCheckSession {
     private outstandingMessages = 0;
     private heartBeatInterval = 3000;
     private iframeRefreshInterval = 60000;
-
-    @Output()
-    onCheckSessionChanged: EventEmitter<any> = new EventEmitter<any>(true);
+    onCheckSessionChanged: BehaviorSubject<any> = new BehaviorSubject<any>(true);
```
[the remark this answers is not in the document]
FabianGosebrink: I agree this should be a Subject. There was a default value before, that's why I thought it would be good to keep it.

profet23:
I agree. I think that's a good change. But you don't want users of the library to be able to emit on the subjects. So you would want to do something like this:

```
private _onConfigurationLoaded = new Subject<boolean>();
public get onConfigurationLoaded(): Observable<boolean> {
    return this._onConfigurationLoaded.asObservable();
}
```

this way users can't call `onConfigurationLoaded.emit(true)` and possibly break internal (and external) things.

src/services/oidc.security.check-session.ts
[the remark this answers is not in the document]
profet23: Yeah I don't think that's a default value. It's being set to async.

src/services/oidc.security.check-session.ts
```diff
@@ -21,7 +21,11 @@ export class OidcSecurityCheckSession {
     private outstandingMessages = 0;
     private heartBeatInterval = 3000;
     private iframeRefreshInterval = 60000;
-    onCheckSessionChanged: BehaviorSubject<any> = new BehaviorSubject<any>(true);
+    private _onCheckSessionChanged = new Subject<any>();
+
+    public get onCheckSessionChanged(): Observable<boolean> {
```
profet23: should be `Observable<any>`

FabianGosebrink:
Thank you for your feedback.

damienbod:
Thank you very much. Doing the testing, and will merge when all is ok

damienbod:
@FabianGosebrink @profet23 Do you think we should release this as 7.0.4 now, or wait?

profet23:
I think this is more of an internal change. 

Externally, users shouldn't see any difference. Probably can wait.

PR #331, opened 2018-10-29
Optional url handler for logoff function

MrGrabazu:
Authorization function has the option to pass a function that hanlde the url instead of redirect to it. I propose that logoff function works the same way and allow the user to pass a function that handle the url as it want or redirect if there is not handler in parameters.


profet23:
Added some unit tests here:  https://github.com/MrGrabazu/angular-auth-oidc-client/pull/1

Mind incorporating those into this pull request?

PR #338, opened 2018-11-07
Update existing-iframe.service.ts

MrGrabazu:
It raise an error on server side rendering because it try to call window despite the fact that it is not existing on server side.

### DIFF
--- a/src/services/existing-iframe.service.ts
+++ b/src/services/existing-iframe.service.ts
@@ -20,7 +20,7 @@ export class IFrameService {
     }
 
     addIFrameToWindowBody(identifier: string) {
-        if(isPlatformBrowser(platformId)) {
+        if(isPlatformBrowser(this.platformId)) {
             const sessionIframe = window.document.createElement('iframe');
             sessionIframe.id = identifier;
             this.loggerService.logDebug(sessionIframe);
@@ -32,14 +32,14 @@ export class IFrameService {
     }
 
     private getIFrameFromParentWindow(identifier: string) {
-        if(isPlatformBrowser(platformId)) {
+        if(isPlatformBrowser(this.platformId)) {
             return window.parent.document.getElementById(identifier);
         }
         return null;
     }
 
     private getIFrameFromWindow(identifier: string) {
-        if(isPlatformBrowser(platformId)) {
+        if(isPlatformBrowser(this.platformId)) {
             return window.document.getElementById(identifier);
         }
         return null;
--- a/src/services/existing-iframe.service.ts
+++ b/src/services/existing-iframe.service.ts
@@ -1,9 +1,13 @@
-import { Injectable } from '@angular/core';
+import { Injectable, Inject, PLATFORM_ID } from '@angular/core';
+import { isPlatformBrowser } from '@angular/common';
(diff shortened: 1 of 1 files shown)


profet23:
While this does fix the error, I think we probably want to disable `silent_renew` and `start_checksession` for server platforms. Since those two operations both require iframes. This would prevent other odd interactions.

Some unit tests wouldn't hurt either.

profet23:
There might also be an argument for saying that the developer should be disabling both the silent renew and check session in their configuration when running on the server. 

Would be curious what @damienbod and @FabianGosebrink think.

damienbod:
@profet23  Yes, when running on the server, the silent renew and check session should be not on.  Just wondering what is the use case for using SSR? I never use this.

profet23:
@damienbod i never use it either. To me it doesn't make sense to use it for client side authorized apps. It may be useful if you want to render unauthorized pages. But yes. I would tend to a "won't fix" and just tell the user to make sure to disable silent renew and check session for SSR.

FabianGosebrink:
I do not use SSR as well but I do see usecases. However we should make it configurable imho.

robisim74:
@damienbod @profet23  @FabianGosebrink SSR does not happen at runtime. With SSR the app _is rendered before_ it is executed, so it is not possible to tell the user to disable an option. The only way is that already followed by this library, and correctly implemented by this pr: use `isPlatformBrowser`

profet23:
@robisim74 what i mean is that the configuration code should utilize `isPlatformBrowser` to determine the correct configuration for the platform.

robisim74:
Ok, the result will be the same: the only important thing is that the code that contains _window_ or manipulates the DOM it is not executed during server-side rendering, and that for this purpose   `isPlatformBrowser` is used.

profet23:
Right. I just think it might be a code "smell" to have the library simply do nothing when it is asked to do a silent renew or a check session routine.

I think I'm of the opinion (although I could be persuaded) that the library is functioning as intended, but the developer has configured it incorrectly for their target platform when its rendering on the server.

robisim74:
I see. But usually that is not left to the developer, but to libraries in order to avoid these problems.

profet23:
If that's convention. I'm fine with it. But I'm not a fan of the approach used in this pull request. 

I think the `AuthConfiguration` class should override `silent_renew` and `start_checksession` to always be false when not running in the browser. 

Again... i could be probably be convinced otherwise.

profet23:
That can go away if we move the check to two properties in the `AuthConfiguration` class.

As it will hit this: https://github.com/damienbod/angular-auth-oidc-client/blob/a1700be90e0f20644e8e3f89f6e2b4cbacd9e5d5/src/services/oidc.security.service.ts#L173

profet23:
@robisim74   Please see my proposal for this here:  #341

robisim74:
Because you merged #341, it should not be necessary.

profet23:
Once #341 is tested/approved we can probably abandon/close this PR.

damienbod:
@MrGrabazu The feature added in this PR is now included in version 8.0.2. #341  Thanks you very much for this PR.

PR #340, opened 2018-11-07
When `logOff()` is called storage should be cleared before emitting a…

### LINKED ISSUE #339: logOff doesn't clear storage
currently the logoff() method isn't clearing storage so the tokens etc. are all still there.

Is there a method that will clear this?  Shouldn't the logOff() method clear these for you?

profet23:
…n authorization event.

If users are redirecting/reloading from within the isAuthorized subscription, storage may not be cleared. This pull request simply changes the order of operations to have storage cleared first, and then emit. 

Unit test added to prevent regression.

Fixes #339


damienbod:
will be released in version 8.0.2

PR #344, opened 2018-11-13
Changed iframe to avoid changing history state for repeated silent token renewals

### LINKED ISSUE #279: Silent Renew adds a state in the parent history in Firefox
It doesn't seem that the issue #203 has been completely fixed.
I tried different version 6.0.3-6.0.6 and it doesn't work in Firefox.
However it works fine in Chrome.

jjasloot:
This pull request should avoid the problem described in #279 by not changing the src but updating the location of the iFrame in another way

Fixes #279


profet23:
Probably need to do this for check session as well.

jjasloot:
@profet23 Included that change as well

Issue #348, opened 2018-11-16
Reset history after reading the tokens in the redirect URL


damienbod:
window.history.replaceState({}, ...

To prevent tokens from being read from the history

damienbod:
https://github.com/damienbod/angular-auth-oidc-client/pull/350

damienbod:
added in version 8.0.3

PR #350, opened 2018-11-20
reset the history

damienbod:
clean up the tokens from the history after a STS callback


damienbod:
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/348

PR #363, opened 2019-01-07
Adding support for Code flow with PKCE

damienbod:
## Details planned released 9.0.0

- OIDC Code Flow with PKCE
- Uses silent renew to get new access tokens
- no support for refresh tokens
- still support for OIDC Implicit Flow

## Breaking changes

Implicit flow callback renamed from authorizedCallback() to authorizedImplicitFlowCallback()

## Docs 
Added to the readme, API docs

## Usage diffs to Implicit Flow

-  response_type :"code"
-  handle callback:

```javascript
private doCallbackLogicIfRequired() {
        console.log(window.location);
        // Will do a callback, if the url has a code and state parameter.
        this.oidcSecurityService.authorizedCallbackWithCode(window.location.toString());
}
```
- silent renew:
```javascript
window.onload = function () {
            /* The parent window hosts the Angular application */
            var parent = window.parent;
            /* Send the id_token information to the oidc message handler */
            var event = new CustomEvent('oidc-silent-renew-message', { detail: window.location });
            parent.dispatchEvent(event);
        };
```

### DIFF
--- a/API_DOCUMENTATION.md
+++ b/API_DOCUMENTATION.md
@@ -62,6 +62,33 @@ Example of a silent_renew.html callback html file.
 
 Note: The CustomEvent does not work for older versions of IE. Add a javascript function instead of this, if required.
 
+### Code Flow with PKCE
+```html
+<!doctype html>
+<html>
+<head>
+    <base href="./">
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>silent-renew</title>
+    <meta http-equiv="content-type" content="text/html; charset=utf-8" />
+</head>
+<body>
+
+    <script>
+        window.onload = function () {
+            /* The parent window hosts the Angular application */
+            var parent = window.parent;
+            /* Send the id_token information to the oidc message handler */
+            var event = new CustomEvent('oidc-silent-renew-message', { detail: window.location });
+            parent.dispatchEvent(event);
+        };
+    </script>
+</body>
+</html>
+
+```
+### Implicit Flow
 ```html
 <!doctype html>
 <html>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 ### 2019-01-08 version 9.0.0
 * Support for OpenID Connect Code Flow with PKCE
 
-Breaking changes:
+### Breaking changes:
+
 Implicit flow callback renamed from authorizedCallback() to authorizedImplicitFlowCallback()
 
 <a name="2018-11-16"></a>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## angular-auth-oidc-client Changelog
 
+<a name="2019-01-08"></a>
+### 2019-01-08 version 9.0.0
+* Support for OpenID Connect Code Flow with PKCE
+
+Breaking changes:
+Implicit flow callback renamed from authorizedCallback() to authorizedImplicitFlowCallback()
+
 <a name="2018-11-16"></a>
 ### 2018-11-16 version 8.0.3
 * Changed iframe to avoid changing history state for repeated silent token renewals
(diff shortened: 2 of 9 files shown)


damienbod:
This is the first implementation. Tested and works. Would be grateful for reviews. I will do the docs if all is ok. The OIDC Authorization Code with PKCE is now recommended from various 'experts' over the OIDC Implicit Flow. I decided to add support for this.

Refresh tokens are NOT supported. The access tokens are refreshed using the iframe like recommended for the Implicit Flow.

codemasterover9000:
Hi, is it possible to hide the details of the response in the library. Specifically: 

```
        const urlParts = window.location.toString().split('?');
        const params = new HttpParams({
            fromString: urlParts[1]
        });
        const code = params.get('code');
        const state = params.get('state');
        const session_state = params.get('session_state');
```

IMO this really belongs to the library. It contains OIDC specific details the user of the library should not be concerned with (parsing of the the response tokens/parameters and checking if it consists of a valid auth response). 

Also for implicit for that matter (user needing to check hash).

Thanks

codemasterover9000:
Hi @damienbod , Thanks for the change. 

I'm not sure how to implement it though in my code. How do I decide to call authorize() or authorizedImplicitFlowCallback()/authorizedCallbackWithCode()?

My original code is like this:
```
if (window.location.hash)
{
    this.oidcSecurityService.authorizedImplicitFlowCallback(window.location.hash);
}
else
{
    this.oidcSecurityService.authorize();
}
```

The internals of authorizedImplicitFlowCallback()/authorizedCallbackWithCode() do not call the authorize() method for me. So I still need to make this decision myself based on the auth response url. Or am I missing something?

I was thinking of a separate method maybe ```this.oidcSecurityService.isAuthResponse(window.localtion)``` for the check.

damienbod:
@codemasterover9000 I need to think about this. For the moment, this could be implemented as follows in 9.0.0

```javascript
const urlParts = window.location.toString().split('?');
const params = new HttpParams({
	fromString: urlParts[1]
});
const code = params.get('code');
const state = params.get('state');
const session_state = params.get('session_state');

if (code && state && session_state) {
	this.oidcSecurityService.requestTokensWithCode(code, state, session_state);
}
else
{
    this.oidcSecurityService.authorize();
}
```

Greetings Damien

codemasterover9000:
Yes I understand. That's why I though a method like ```isAuthResponse(Location | string)``` would still allow that kind of freedom to the end user of the library.

To be used like this:
```
if (this.oidcSecurityService.isAuthResponse(window.location))
{
    this.oidcSecurityService.authorizedImplicitFlowCallback(window.location.hash);
}
else
{
    this.oidcSecurityService.authorize();
}
```
The method checks if the url pattern matches on of the supported flows. It is still up to the user to call the ```authorizedImplicitFlowCallback()``` or ```authorizedCallbackWithCode()``` methods then depending on their case.

KylerJohnson26:
Does anyone have a sample `app.module.ts` or a sample Angular project I can look at which is set up for the code flow with PKCE? I'm looking at the documentation and I'm calling `authorizedCallbackWithCode()` with the code and state parameters present in my url, but nothing is happening. I must be missing something in the configuration.

damienbod:
https://github.com/damienbod/angular-auth-oidc-client/blob/master/projects/sample-code-flow/src/app/app.component.ts#L49

KylerJohnson26:
> https://github.com/damienbod/angular-auth-oidc-client/blob/master/projects/sample-code-flow/src/app/app.component.ts#L49

@damienbod 
In this example in the `app.module.ts`, the subscribe method parameter is of type `ConfigResult`, but I get an error saying that angular-auth-oidc-client has not exported member named `ConfigResult`. I'm having the same problem with the type `OpenIdConfiguration` for the config object.

In NPM, I see that the latest version is 10.0.2, but when I run `npm i -s angular-auth-oidc-client`, it installs version 9.0.8. Could that be the reason for this issue? Is it recommended to specify version 10.0.2 explicitly in my package.json?

Thanks so much for you help, by the way!

PR #369, opened 2019-01-11
Turn off history reset per config

damienbod:


csourcenl:
Perfect! This will help me a lot. How can I buy you a beer?

damienbod:
@csourcenl no problem, thanks for the support, that's enough, no beer is required.

cheers Damien

PR #395, opened 2019-02-22
use !! to display getIdToken() and _userData.value in silentRenewHear…

l1b3r:
…tBeatCheck()

These properties are string-based and eventually both stored in the localStorage/sessionStorage. When they are cleared in the other parts of the library, an empty string is set instead, which later leads to an incorrect debug logging in the console if it's being compared against `null` (as of now):

```
> window.localStorage.setItem('someKey', '')
> window.localStorage.getItem('someKey') != null
true
> !!window.localStorage.getItem('someKey')
false
```


damienbod:
Release version 9.0.3

PR #403, opened 2019-03-20
moving to angular package format with cli

FabianGosebrink:


package.json
```diff
@@ -1,102 +1,53 @@
...
-        "oidc",
-        "oauth2",
-        "openid",
-        "security",
-        "typescript"
-    ],
-    "license": "MIT",
+    "private": true,
     "dependencies": {
+        "@angular/animations": "~7.2.0",
```
damienbod: we supported Angular 6 up to now, need to know if this is a breaking change

projects/angular-auth-oidc-client/README.md
```diff
@@ -0,0 +1,24 @@
+# AngularAuthOidcClient
+
```
damienbod: we need to use the existing readme with the build changes

package.json
FabianGosebrink: This is not the lib, this is the project. The peerdependency describes what we support.

PR #428, opened 2019-04-13
Update app.module.ts

vit100:
Changed to method load_using_stsServer().

Also who is owner of https://offeringsolutions-sts.azurewebsites.net pls add http://localhost:4200 in allowed callbacks, as with this sample auth server rejects auth request with message:
>Sign in
>Sorry, but we’re having trouble signing you in.

>AADSTS50011: The reply url specified in the request does not match the reply urls configured for the application: '99eb0b9d-ca40-476e-b5ac-6f4c32bfb530'.

### DIFF
--- a/projects/sample-code-flow/src/app/app.module.ts
+++ b/projects/sample-code-flow/src/app/app.module.ts
@@ -12,7 +12,7 @@ import {
 import { AppComponent } from './app.component';
 
 export function loadConfig(oidcConfigService: OidcConfigService) {
-    return () => oidcConfigService.loadUsingStsServer('https://offeringsolutions-sts.azurewebsites.net');
+    return () => oidcConfigService.load_using_stsServer('https://offeringsolutions-sts.azurewebsites.net');
 }
 
 @NgModule({


FabianGosebrink:
Yes, we should replace the url. However, the method `load_using_stsServer` is going to be replaced, as it is convention to write methods in Javascript without the underscore. However for backwards compatibility both methods are still allowed. Documentation should show the newer one. What do you think @damienbod ?

vit100:
Unfortunately method with underscore gives error




On Sat, Apr 13, 2019 at 12:46 PM -0400, "Fabian Gosebrink" <notifications@github.com> wrote:












Yes, we should replace the url. However, the method load_using_stsServer is going to be replaced, as it is convention to write methods in Javascript without the underscore. However for backwards compatibility both methods are still allowed. Documentation should show the newer one. What do you think @damienbod ?

FabianGosebrink:
So when the method `load_using_stsServer` gives an error we have to fix that. Let me have a look at that. This PR uses the method explicitly, is that intended?

damienbod:
everything is working now, _ method works

PR #474, opened 2019-07-22
Allow the use of Refresh Codes

lthorne1:
We don't want to mess with special headers or silent renew iframes to get new tokens so this method allows us to use the refresh code to obtain a new access token. It operates on the 'code' flow and preempts the silent-renew process if the refresh code exists.


damienbod:
@lthorne1 thanks for the PR. I will go throught the PR and test.

vadjs:
@damienbod Did you test it already?

damienbod:
no, going through it now, wasn't sure if this was a good idea or not.

damienbod:
@lthorne1 thanks for the PR, will release (10.0.4) this hopefully this week, want to add some seperate doc improvement first.

PR #499, opened 2019-09-11
Ability to change the amount of seconds for the IsAuthorizedRace to do a Timeout

juanp3053:
I've been having an issue that if I run the STS server locally or through a very slow connection (I have no other option) the request sometimes takes longer than 5 seconds (I know it isn't normal). Then, if the value for “IsAuthorizedRace” Timeout is hard-coded I do not have the ability to change it, so that my "slower" request can retrieve the Token and Log In to my applications. Adding a configurable parameters maintain the default value of 5 seconds, would help for people with those extreme cases.

Thanks.

### DIFF
--- a/API_DOCUMENTATION.md
+++ b/API_DOCUMENTATION.md
@@ -166,6 +166,10 @@ If this is active, the history is not cleaned up at an authorize callback. This 
 
 default value : 3
 
+### isauthorizedrace_timeout_in_seconds
+
+default value : 5
+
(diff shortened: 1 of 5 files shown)


damienbod:
@juanp3053 

thanks, nice feature, this has been requested a few times. Could you fix the tests? Would be cool

Thanks Damien

juanp3053:
@damienbod 

Done!

Thank you

damienbod:
thanks!!

Issue #479, opened 2019-07-26
RefreshSession and Authorize can cause a nonce validation error


Cracien:
Library Version: 9.0.8
Angular Version: 6.1.0
Browsers Tested: Chrome (72) / IE 11
Description: If you call OidcSecurityService.authorize() from inside of the OidcSecurityService.onAuthorizationResult subscription you will cause the application to fail the nonce validation after being redirected from your IDP.

Requirements for exhibited behavior:
-  Silent Renew is enabled
- You must not have an active session with the IDP

Assumptions
- The IDP returns a login_required if there is no active session and a prompt=none request is made

Repro Steps:

1. Subscribe to the onAuthorizationResult Observable
2. Inside of the subscription do the following:
   a. If the validation result is LoginRequired then call authorize

Expected Behavior
- The app generates / stores a nonce and state in session storage
- The app generates the IDP url based on nonce / state
- The browser is redirected to the IDP url and logs in / is redirected to app
- The app validated the token using the nonce and state values stored in session state

Actual Behavior
- The app generates / stores a nonce and state in session storage
- The app generates the IDP url based on nonce / state
- The app clears the nonce
- The browser is redirected to the IDP url and logs in / is redirected to app
- The app fails to validated the token using the nonce (stored nonce is empty)

Cause
This is caused for two reasons in the oidc.security.service.ts file
1) The authorizedCallbackProcedure clears the nonce after calling the onAuthorizationResult. (Lines 450/456)
2) The browser finishes executing the call chain prior to redirecting the browser (Line 428 is executed but does not actually redirect until after the call chain finishes executing)

The outcome is that the nonce is updated as a result of Line 450 executing (because we are calling the authorize from within the subscription) and is subsequently cleared when line 456 is executed. When the application receives the tokens it no longer has the nonce and results in a validation error.

damienbod:
closing this, please open if you still have an issue

PR #501, opened 2019-09-16
Do not check idToken nonce when using refreshToken

### LINKED ISSUE #491: Refresh Token Flow not working
Hey,

I'm trying to use the new refresh token, but...

If the refresh response includes the id_token, the validation fails with a nonce mismatch error,
If the response doesn't includes the id_token, the validation fails because the sub in the local id_token (now empty) does not match the one in the userinfo endpoint

WoMayr:
Should fix #491 (and maybe #479)

This PR disables the nonce check when receiving an authorization code response after requesting a token with a refresh token as these do not contain a nonce.


WoMayr:
I don't think the build failed because of something I changed? @damienbod could you please take a look at it?

damienbod:
@WoMayr, I will go through this, hopefully at the weekend, looks good and I should be able to merge, release it. Thanks for your patience.

Greetings Damien

PR #510, opened 2019-10-04
OidcSecurityService with some observables

timo-haas:
I added an Observable alternative (suffixed with $) for following methods authorizedCallbackWithCode, requestTokensWithCode, requestTokensWithCodeProcedure

Observables will throw an error result on some occasions, e.g. network timeout on token request
Without return the observables one could not react to these error, e.g. retry the token request


damienbod:
@timo-haas  thanks, looks good, will go through this, the would be a big improvement, thanks

Greetings Damien

damienbod:
will be released with 10.0.9, need to do some testing first, but should be able to release this weekend

Thanks Damien

PR #515, opened 2019-10-05
fixed possible memory leak in silent renew

timo-haas:
- IFrameService
  - only return typesafe HTMLIFrameElement
---
- OidcSecuritySilentRenew
  - made service stateless   
  - fix: sessionIframe.contentWindow could sometimes be null (if DOM has yet to finish loading iframe element), replaced be setting sessionIframe.src attribute
  - fix: removed possible memory leak in sessionIframe.onload (never removed listener)
---
- OidcSecurityService
  - fix: not all usage of refreshSession subscribed
  - line 176; this.oidcSecuritySilentRenew.initRenew(); seems kind of obsolete now?


damienbod:
looks good, thanks! I will test this, hopefullly today

PR #518, opened 2019-10-13
ignore errors on non-observables

timo-haas:
An .subscribe() without arguments will implicitly throw errors if there was an error in the observable.
So I added .subscribe(() => {}, () => {}); for ignoring such cases.

Readded throwError in authorizedCallbackWithCode$(). Which are now ignored in authorizedCallbackWithCode()


damienbod:
@timo-haas Thanks for the PR, sorry that we could  not merge this. We have done a major refactor in the version for Release 11, and cannot merge this anymore.

Thanks Damien

PR #529, opened 2019-11-09
Added renew process denotation to AuthorizationResult

JerrettDavis:
When subscribing to `onAuthorizationResult`, there is no convenient way to determine if the AuthorizationResult is the result of the auto-renewal process or if it comes from an explicit login event. Most of the various handlers in the `OidcSecurityService` already had some knowledge as to whether the source of the event was from the auto-renewal process. So I added a new boolean field, `isRenewProcess`, to AuthorizationResult and updated the subject updates to include the new field.


rekosko:
When can we expect this PR to be merged?

damienbod:
@rekosko @JerrettDavis hopefully this week

Issue #519, opened 2019-10-13
silent_renew bug


klanglie:
With silent_renew enabled, refreshSession() / createAuthorizeUrl is being called prior to authorize(urlHandler). This leads to what may be a race condition. Expected behavior would be to block refreshSession() until authorize is complete. It might be something IDSV4-related. This didn't seem to happen before the upgrade to v3.

{
  "stsServer": "https://localhost:44318",
  "redirect_url": "http://localhost:4400/oidc-callback",
  "client_id": "myclient",
  "response_type": "code",
  "scope": "mySiteAccess openid profile",
  "post_logout_redirect_uri": "http://localhost:4400",
  "start_checksession": true,
  "silent_renew": true,
  "silent_renew_offset_in_seconds": 0,
  "silent_renew_url": "http://localhost:4400/silent-renew.html",
  "use_refresh_token": false,
  "post_login_route": "/home",
  "forbidden_route": "/forbidden",
  "unauthorized_route": "/unauthorized",
  "log_console_warning_active": true,
  "log_console_debug_active": false,
  "max_id_token_iat_offset_allowed_in_seconds": 120,
  "apiServer": "https://localhost:44390/",
  "apiFileServer": "https://localhost:44378/",
  "siteName": "My Site"
}

Error: Invalid redirect_uri{ redirectUri: "http://localhost:4400/silent-renew.html", expectedRedirectUri: "http://localhost:4400/oidc-callback" }, details: TokenRequestValidationLog { ClientId: "myclient", ClientName: "myclient", GrantType: "authorization_code", Scopes: null, AuthorizationCode: "i0-mw8cjUT68Z-JajYmzze5DEFVxG-_4QAWdDQYc3xA", RefreshToken: null, UserName: null, AuthenticationContextReferenceClasses: null, Tenant: null, IdP: null, Raw: [("grant_type": "authorization_code"), ("client_id": "myclient"), ("code_verifier": "C0.30956101666064106157083750138015708375013800.8447167887744094"), ("code": "i0-mw8cjUT68Z-JajYmzze5DEFVxG-_4QAWdDQYc3xA"), ("redirect_uri": "http://localhost:4400/silent-renew.html")] }

FabianGosebrink:
Will be fixed in rls 11 as well

BrettJG:
@FabianGosebrink , any ETA on release 11?

FabianGosebrink:
[...]

Issue #498, opened 2019-09-11
Bad request when calling authorizedCallbackWithCode()


Maikell84:
I have a problem implementing the library and I'm not sure if it is a bug or if I'm doing it wrong.
I am using **Angular 8.2.4** on client side with version 10.0.6 of this library , and an **Identity Server 4** with code-flow as authorization server.

After successfully logging in at the server, I get correctly redirected to my configured angular-route. Here I have created a component, which handles the callback. This unfortunately is not working.

```
// Redirect Component
public ngOnInit(): void {
    this.authorizationService
        .getOidcClientReady()
        .pipe(
            take(1),
            switchMap(() => {
                return this.oidcSecurityService.getIsAuthorized();
            }),
            distinctUntilChanged(),
            filter(isAuthorized => !isAuthorized),
            takeUntil(this.isDestroyed$),
        )
        .subscribe(() => {
            // If we are not authorized, we must call authorizedCallbackWithCode()-function with code, state and session_state params from url.
            // If params are valid, onAuthorizationResult-Observable will fire, where we perform the loading of the user-object
            this.oidcSecurityService.authorizedCallbackWithCode(window.location.toString());
        });
}
```

`getOidcClientReady()` is an own function that emits, as soon as the library is setup and the configuration is loaded.

My app module looks like this:

```
export class AppModule {
    constructor(private oidcSecurityService: OidcSecurityService, private oidcConfigService: OidcConfigService, private router: Router) {
        this.oidcConfigService.onConfigurationLoaded.subscribe((configResult: ConfigResult) => {
            const config: OpenIdConfiguration = {
                stsServer: configResult.customConfig.stsServer,
                redirect_url: window.location.origin + '/admin/#/redirectLogin',
                client_id: 'myClientId',
                scope: 'openid profile email IdentityServerApi offline_access',
                response_type: 'code',
                silent_renew: true,
                storage: localStorage,
                silent_renew_url: window.location.origin + '/admin/assets/staticSites/silent-renew.html', // Static iframe, that handles silent refresh events
                log_console_warning_active: true, 
                log_console_debug_active: true,
                auto_userinfo: true,
                trigger_authorization_result_event: true, 
            };

            try {
                this.oidcSecurityService.setupModule(config, configResult.authWellknownEndpoints);
            } catch (error) {
                this.router.navigate(['/notAuthorized'], { queryParams: { error: error } });
            }
        });
    }
}
```

[...]

Issue #476, opened 2019-07-25
redirect_url mismatch


twmobius:
First off let me thank you for your time building and maintaining this module. 

I've been using implicit flow with `silent_renew: false` for some time now but recently I've tried to change the client to use `code PCKE flow` with `silent_renew: true` but it seems that something is wrong.

Here is what I'm doing. 

First off this is the configuration used:

```
export function loadConfig(oidcConfigService: OidcConfigService) {
  return () => oidcConfigService.load_using_stsServer(environment.server);
}

@NgModule({
  imports: [
    SharedModule,
    RouterModule.forChild([
      {
        path: 'login',
        component: LoginComponent
      },
      {
        path: 'login/unauthorized',
        component: UnauthorizedComponent
      },
      {
        path: 'login/forbidden',
        component: ForbiddenComponent
      },
      {
        path: '',
        component: LoginComponent
      }
    ]),
    AuthModule.forRoot(),
  ],
  declarations: [
    LoginComponent,
    UnauthorizedComponent,
    ForbiddenComponent
  ],
  providers: [
    OidcConfigService,
    {
      provide: APP_INITIALIZER,
      useFactory: loadConfig,
      deps: [OidcConfigService],
      multi: true,
    },
  ],
})
export class AuthenticationModule {
  constructor(
    private oidcSecurityService: OidcSecurityService,
    private oidcConfigService: OidcConfigService,
  ) {
    const location = window.location;

    this.oidcConfigService.onConfigurationLoaded.subscribe((configResult: ConfigResult) =>{
      const config: OpenIdConfiguration = {
        stsServer: environment.server,
        redirect_url: 'http://localhost:4200/login',

        client_id: environment.clientId,
        response_type: 'code',
        scope: 'openid email profile telephone address',

        post_logout_redirect_uri: `${location.protocol}//${location.hostname}/login`,
        forbidden_route: '/login/unauthorized',
        unauthorized_route: '/login/unauthorized',

        start_checksession: false,
        silent_renew: true,
        silent_renew_url: `${location.protocol}//${location.hostname}/assets/silent-renew.html`,

        log_console_warning_active: true,
        log_console_debug_active : true,
        max_id_token_iat_offset_allowed_in_seconds : 100
      };

      this.oidcSecurityService.setupModule(config, configResult.authWellknownEndpoints);
    });
  }
}

```

I've build a AuthGuard that checks if the user is authenticated and if not it redirects to route `/login`. 

[...]

PR #534, opened 2019-11-22
PCKE proper access token request.

elwray:
According to RFC 7636 (4.5) and RFC 6749 (4.1.3): "if the "redirect_uri" parameter was included in the authorization request as described in Section 4.1.1, and their values MUST be identical."

redirect_uri should be the same for authorization request and for the access token request!

RFC 7636 (4.5): https://tools.ietf.org/html/rfc7636#section-4.5
RFC 6749 (4.1.3): https://tools.ietf.org/html/rfc6749#section-4.1.3

Should fix the next bug: https://github.com/damienbod/angular-auth-oidc-client/issues/498
And looks like should fix: https://github.com/damienbod/angular-auth-oidc-client/issues/519 and https://github.com/damienbod/angular-auth-oidc-client/issues/476

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/services/oidc.security.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/services/oidc.security.service.ts
@@ -413,17 +413,12 @@ export class OidcSecurityService {
             return throwError(new Error('incorrect state'));
         }
 
-        let headers: HttpHeaders = new HttpHeaders();
-        headers = headers.set('Content-Type', 'application/x-www-form-urlencoded');
+        const headers: HttpHeaders = new HttpHeaders()
+          .set('Content-Type', 'application/x-www-form-urlencoded');
 
-        let data =
+        const data =
             `grant_type=authorization_code&client_id=${this.configurationProvider.openIDConfiguration.client_id}` +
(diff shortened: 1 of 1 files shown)


elwray:
@damienbod can you please review my PR?

damienbod:
Hi @elwray 

sorry for the delay, very busy at the day job, will get back to you. Thanks for the PR!

Greetings Damien

damienbod:
Hi @elwray 

This is required for the silent running. When the silent renew sends an authorize request it is correct, so this code is required.

Greetings Damien

elwray:
Hi @damienbod 

When we generate a link for authorization request we pass "redirect_url" from config. So, for access token request we SHOULD use the same redirect_url, but we instead of same redirect_url we use "silent_renew_url".

damienbod:
Hi @elwray I validated this in some of the implementations and the lib works as per spec. When the authorize is sent for the silent renew, the renew URL is used for both requests, and the login authorize uses the redirect URL for both. 

Thanks for your feedback.

Greetings Damien

damienbod:
@elwray Thanks for the PR, sorry that we could  not merge this. We have done a major refactor in the version for Release 11, and cannot merge this anymore.

Thanks Damien

PR #551, opened 2020-01-07
adding a property to the OpenIdConfiguration model to allow for a cus…

JoellyR:
custom discovery url, along with a change in the security.config service on line 68 to add a check to see if a custom url exists and if not to fall back to the default.

This is in regards to issue 539... The case for a custom discovery url is sometimes asked for in such cases as the following: http://www.hl7.org/fhir/smart-app-launch/#for-example in section 6.1.1 they refer to the meta data url (pretty much the discovery url) as  .well-known/smart-configuration.json instead of  .well-known/openid-configuration.json and to support these cases is why I think this pull request should be merged into the master branch


damienbod:
@JoellyR Thanks for the PR, sorry that we could  not merge this. We have done a major refactor in version for Release 11, and cannot merge this anymore.

Thanks Damien

PR #579, opened 2020-02-11
subscribe startRenew after isAuthorized is true

gjberghuis:
When silentRefresh enabled with code pkce flow results in a first call to get a token ('/connect/token') before the authorization call has been finished ('/authorize' with prompt none). The token calls after completion of the authorize call are succesful but the call before results in a '400' error.

In my usecase I wanted only to start with silentRefresh after I've been authorized. 

This refers to the same issue here  -> https://github.com/damienbod/angular-auth-oidc-client/issues/519


damienbod:
This looks good, build fails to node 11 in the build config which I updated. I will merge.

Thanks

damienbod:
build works now, I will do some manual testing, and create a new npm package when this is good

Thanks for the fix!

Greetings Damien

gjberghuis:
Thnx Damien! 

I'll be waiting for the new npm package. 

Greetings

Gert-Jan

damienbod:
released in 10.0.15
thanks

alecrt:
This pull request prevent initial silent login.
See issue  #583

PR #616, opened 2020-03-27
Add Audience to generated login URL

leviplj:


damienbod:
@leviplj Thanks for the PR.

just wondering, why do you need this in the URL?

damienbod:
Hi @leviplj 

The audience is not an standard parameter for the OIDC request. You can add any custom parameters you want by using the customParams function. 

thanks for the PR.

Greetings Damien

PR #670, opened 2020-04-25
Extract small functions

damienbod:


projects/angular-auth-oidc-client/src/lib/authState/auth-state.service.ts
```diff
@@ -121,7 +121,7 @@ export class AuthStateService {
         this.storagePersistanceService.authorizedState = authState;
     }
 
-    private tokenIsExpired() {
+    public tokenIsExpired() {
```
FabianGosebrink, line 124: dont write the `public` keyword.

PR #678, opened 2020-04-27
validate if access token expired

damienbod:
When the access token has been returned, the expires_in property is used to define when this expires. This is then used to validate if the access token has expired


projects/angular-auth-oidc-client/src/lib/authState/auth-state.service.ts
```diff
@@ -100,11 +100,14 @@ export class AuthStateService {
...
-            this.loggerService.logDebug('persisted token is expired');
+        if (this.hasIdTokenExpired()) {
+            this.loggerService.logDebug('persisted id_token is expired');
+            return false;
+        } else if (this.hasAccessTokenExpiredIfExpiryExists()) {
+            this.loggerService.logDebug('persisted access_token is expired');
             return false;
         } else {
-            this.loggerService.logDebug('persisted token is valid');
+            this.loggerService.logDebug('persisted id_token and access token are valid');
```
FabianGosebrink, line 110: remove the `else` keyword for clarity. As we work with early returns we do not need the else.

damienbod: fixed

PR #682, opened 2020-04-27
remove isauthorized race timeout in seconds

damienbod:


.gitignore
```diff
@@ -46,3 +46,5 @@ testem.log
 Thumbs.db
 
 /.angulardoc.json
+/projects/angular-auth-oidc-client/LICENSE
```
FabianGosebrink, line 49: Do not modify this as when the files are getting copied in this folder per accident you see it in the commits and do not drive on with those wrong changes

PR #696, opened 2020-04-30
Fabiangosebrink/adding unit tests

FabianGosebrink:


projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -99,36 +103,36 @@ export class UrlService {
         const codeVerifier = this.flowsDataService.getCodeVerifier();
         if (!codeVerifier) {
             this.loggerService.logWarning(`CodeVerifier is not set `, codeVerifier);
+            return null;
         }
 
-        let data = oneLineTrim`grant_type=authorization_code&client_id=${this.configurationProvider.openIDConfiguration.clientId}
+        const url = oneLineTrim`grant_type=authorization_code
```
damienbod, line 109: should be body

Issue #721, opened 2020-05-10
athash is not returned in code-flow by Azure B2C failing validation.


msdate:
Testing B2C + SPA with code flow, we are noticing that the server issues a valid id_token, however will at_hash is null.  As per, https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens, The access token hash is included in ID tokens only when the ID token is issued with an OAuth 2.0 access token. 

The at_hash is optional for the code flow, however if its missing, it triggers => handleUnsuccessfulValidation

Not sure if this is an issue or what the fix is, i attempted a PR. https://github.com/damienbod/angular-auth-oidc-client/pull/720

![B2C](https://user-images.githubusercontent.com/12368016/81511895-9bac8a00-92ea-11ea-8a10-80372a79e4b6.png)

damienbod:
@msdate Thanks for reporting. I will validate this. We have with other STS as well. Does it work now with your fix, from the PR?

I will update the B2C sample to code flow then, if the code flow is supported and works.

Thanks Damien

damienbod:
@msdate I tested this, and had to add a small change to make B2C work. Have everything now working in a PR. Testing and validating now, and will release this in 11.0.2

msdate:
For Azure B2C, we made the following changes to manifest, and added type Spa, this option will not need to configure secrets and is working.  
The configured options that work for us.
return oidcConfigService.withConfig({ 
......
          scope: 'openid',
          responseType: 'code',
          autoUserinfo: false
....})
"replyUrlsWithType": [
		{
			"url": "http://localhost:4200/",
			**"type": "Spa"**
		}
	]

damienbod:
I had to add another option to deactivate the auth_time in the id_token of the refresh session. Azure B2C have not implemented the spec correctly. I just used a native client as the app type, and you don't require a secret either. I have a refresh session using refresh tokens.

msdate:
We had a question on the 3 sec check for the refresh token, does the above fix eliminate these? Also how do the refresh token reissue work, is this polled on a fixed duration for evaluation, or it is triggered when the token is due to expire (approaching expiry) rather than on a polling basis?

![image](https://user-images.githubusercontent.com/12368016/81589832-d44e7100-9387-11ea-80da-bec297f4b69b.png)

Thanks again for the quick fix on this issue.

damienbod:
Yes the refresh works with the fix. 

[...]

PR #722, opened 2020-05-11
Azure B2C Code Flow PKCE support

damienbod:


damienbod:
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/721

Issue #729, opened 2020-05-13
V11 Force Get New Token


JohnGalt1717:
**Describe the bug**
In v10 you could call refereshSession() which would cause a new token to be retrieved (access and id if you specified that)  V11 doesn't appear to have this.

**Expected behavior**
There still needs to be a function for the client to say "Hey, I know something changed with my token, give me a new one."

damienbod:
@JohnGalt1717 This would be good, sorry about this getting removed.

JohnGalt1717:
Np! For now I'm just calling login again, which causes the idp to force a new token loop, but obviously that's not very efficient. Would be great if it was just an ajax call again.

damienbod:
@JohnGalt1717  Just realized, this should work for you in the same way:

```
refresh() {
    this.oidcSecurityService.authorize({ customParams: { prompt: 'none' });
}
```

If this would work, then we could save adding the new API. Would this work for you?

Greetings Damien

JohnGalt1717:
So my issue with that is that's basically what I'm doing right now. The problem is that it's looping through the idP.  I would really like it to not have to reconstruct the angular site as a result and just do an ajax call to get a new auth and id token on demand and update everything in it so that I can update the interface and security rules etc. without a reload.

Before I could use refreshSession() which did this without a loop through and thus tear down and rebuild of the angular site which was ideal.

damienbod:
yes, good idea. Ok I will add a forceRefreshSession() API which will do this

JohnGalt1717:
Thanks!

damienbod:
Added this to version 11.1.0, should be releasing soon, just what to do some tests tomrrow, thanks for reporting.

Greetings Damien

JohnGalt1717:
You rock! Thanks!

teebszet:
Hi! 

Using this feature with implicit flow, and I run into a problem where redirect_uri query param is not being set on the call to /authorize.

I do have redirectUrl set on my startup config. 

An example request that the forceRefreshSession() tries to make is:

[...]

PR #730, opened 2020-05-13
force refresh session, new API

damienbod:


damienbod:
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/729

JohnGalt1717:
I just tried this and it doesn't appear to be reloading the Id Token even though my setup is set to auto-load the id.

damienbod:
It's an async method, just in case, add a subscribe to force the call?

I use it like this:

https://github.com/damienbod/angular-auth-oidc-client/blob/master/projects/sample-code-flow-refresh-tokens/src/app/home/home.component.ts#L26-L28

Can you compare? Also you can check in the F12. Just wondering, do you have the silent renew active?

Greetings Damien

JohnGalt1717:
Here's how I'm calling it: 

const result = await this.oidcSecurityService.forceRefreshSession().toPromise();

result = true;

No event fired for the auth change, and no update of the id_token.

Here's my config:
		return () =>
			oidcConfigService.withConfig({
				stsServer: apiUrl,
				redirectUrl: `${location.protocol}//${location.host}`,
				postLogoutRedirectUri: LOGOUT_PATH,
				postLoginRoute: DEFAULT_LOGIN_REDIRECT_PATH,
				clientId: CLIENT_ID,
				scope: 'openid profile email',
				responseType: 'code',
				silentRenew: true,
				storage: localStorage,
				silentRenewUrl: `${location.protocol}//${location.host}/silent-renew.html`,
				logLevel: environment.production ? LogLevel.Error : LogLevel.Debug,
				autoUserinfo: true,
				forbiddenRoute: FORBIDDEN_PATH,
				autoCleanStateAfterAuthentication: true,
				triggerAuthorizationResultEvent: true,
				maxIdTokenIatOffsetAllowedInSeconds: 600,
				unauthorizedRoute: UNAUTHORIZED_PATH,
				renewTimeBeforeTokenExpiresInSeconds: 20,
			});


I see the silent-renew.html
https://localhost:5002/silent-renew.html?code=<code here>

Oddly, if I hit F5, I do get the updated id_token information from getPayloadFromIdToken(), it just isn't updated after the forceRefreshSession() is called;

damienbod:
Hi @JohnGalt1717 Your code looks good. would it be possible to supply an example?

Greetings Damien

JohnGalt1717:
If you take the link you provided above and replace console.log with:

const userData = this.oidcSecurityService.getPayloadFromIdToken();
console.log (userData);

You'll see that any changes to the id_token won't be reflected with a call from getPayloadFromIdToken(), but if you look in localStorage and decode the id_token in xxx_authorizationDataIdToken it does have it.

Here's the general flow:

1. Update some field via api that the idp encodes in the id_token
2. Call forceRefreshSession()
3. UI updates based on new values from the idp in the id_token from call to getPayloadFromIdToken that sets fields.

Preferably it would be:

1. Update some field
2. forceRefreshSession
3. checkAuth involked automatically by this library because the auth changed because the tokens changed.
4. Respond to event like normal which would cause the right stuff to be updated which would change the UI.

So I'd say that the bug is that getIdFromToken still has a cached copy even though it was updated and stored elsewhere so it never gets the new version, and the forceRefreshSession doesn't compare the auth and id tokens it gets back and then call checkAuth if they aren't the same. (doesn't matter about the content, only that they're not identical)

damienbod:
Thanks
I'll create a new issue with this the validate./solve Thanks for this info.

JohnGalt1717:
Note, I think this is a timing issue caused by something that is observable/async not being awaited. If I put a breakpoint after the forceRefreshSession and wait a few seconds, and then use getIdFromToken it is updated.

damienbod:
@JohnGalt1717 moved the issue here:

https://github.com/damienbod/angular-auth-oidc-client/issues/734

I think you just need to wait for it to come back. We have a notification event.

Issue #725, opened 2020-05-13
Obtaining new credentials with refresh code failed


Kayes-Islam:
**Describe the bug**

**To Reproduce**
My configuration is:
`      oidcConfigService.withConfig({
          stsServer: 'https://our-server.com/oauth',
          redirectUrl: window.location.origin,
          postLogoutRedirectUri: window.location.origin,
          clientId: 'our-app',
          scope: 'openid profile ourapi',
          responseType: 'code',
          silentRenew: true,
          useRefreshToken: true,
          logLevel: LogLevel.Debug
      });`

We are using identity 4 server with one time only refresh token, below is the configuration:
![image](https://user-images.githubusercontent.com/9096105/81765396-db55ba80-9512-11ea-8860-1491abc46ca1.png)

Here's the browser log of failure:
```
IsAuthorized: id_token idTokenHasExpired, start silent renew if active
BEGIN refresh session Authorize
RefreshSession created. adding myautostate: _some_code_here_
found refresh code, obtaining new credentials with refresh code
POST https://beta-rest.oceaninformatics.com/oauth/connect/token 400 (Bad Request)
```
Looking at the browser's network tab the response to the "token" endpoint returned 404 with body: `{"error":"invalid_grant"}`

Interestingly, probably the source of the problem, the form data from from the request shows token as undefined. Here's the raw of the request form data form network tab:
`grant_type=refresh_token&client_id=our-app&refresh_token=undefined`

**Expected behavior**
Silent renew with refresh token should work as expected.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version  81.0.4044.129

Kayes-Islam:
Doing a bit more debug looks like there's an issue with AuthStateService.getRefreshToken():
```
        const token = this.storagePersistanceService.getRefreshToken();
        return decodeURIComponent(token);
```
If token is undefined decode turns it into a string "undefined". However, I'm still not sure why the refresh token is not there in authResult.

Kayes-Islam:
Looking at OpenID connect Core, the client must include offline_access scope. So I changed my configuration to include offline_access in the scope: 
`scope: 'openid profile ourapi offline_access'`

This has fixed the issue on my side. However, I think AuthStateService.getRefreshToken() can be improved as per comment above and this card can be closed.

damienbod:
[...]

PR #731, opened 2020-05-14
Fabiangosebrink/add config validation

FabianGosebrink:
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/725


FabianGosebrink:
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/734

Issue #744, opened 2020-05-18
Silent renew iframe is not loaded when there is no session


GabrielGil:
**Describe the bug**
Silent renew, should (and it used to), be triggered when enabled, so the application can be authorised directly without redirects in case the user is already authenticated in the STS server.

**To Reproduce**
Steps to reproduce the behaviour:
1. Configure a project using the sample: https://github.com/damienbod/angular-auth-oidc-client/blob/master/docs/samples.md#code-flow-with-pkce-basic-with-silent-renew
2. Open the application in *Tab 1*
3. Login (At this point, the silent renew iframe is loaded and working)
4. Open the application in *Tab 2*
5. The silent renew iframe is not loaded, forcing the user to click on "Login" again.

**Expected behaviour**
From the previous steps, I believe the plugins should load the iframe to silently renew. This will cause the STS to give the app new tokens directly, as the app was previously authorised already.

This is the behaviour of the silent renew as per version `10.0.14`.

Also, I'd like to clarify that this is not auto-login, given that auto-login will redirect the user to the STS login interaction, regardless of the current user state in the STS whereas silent renew will only get a fresh set of tokens in case the user is still correctly authenticated and the client app authorised in the STS server.

**Screenshots**
![image](https://user-images.githubusercontent.com/961534/82215215-5229f380-9917-11ea-9965-0dc7efd006ed.png)

**Desktop (please complete the following information):**
 - OS: Mac OS 10.15.4
 - Browser: Chrome, Firefox, Edge, Safari.
 - Version 81, 75, 84, 13.1

**Smartphone N/A:**

**Additional context**
N/A

damienbod:
Hi @GabrielGil yes we changed the startup here. We wanted to simplify the logic, and removed all our timing issues. You can achieve the same with the following code:

```
 ngOnInit() {
        this.oidcSecurityService.checkAuth().subscribe((isAuthenticated) => {
            console.log('app authenticated', isAuthenticated);
            if (!isAuthenticated) {
                console.log('run login without UI', isAuthenticated);
                this.oidcSecurityService.authorize({ customParams: { prompt: 'none' } });
            }
        });
    }
```

Then it will work like you want.

Hope this helps Damien

GabrielGil:
Hi @damienbod.  Thanks for your answer.

That is not exactly precise, as it is not the same result and it has a number of side effects:
[...]

PR #752, opened 2020-05-21
server check silent renew

damienbod:


FabianGosebrink:
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/744

PR #761, opened 2020-05-28
Refresh checksession iframe regularly

### LINKED ISSUE #756: Check session always checked locally.
Hello.
In the recursive function doesn't include call to CheckSessionService.init() :

https://github.com/damienbod/angular-auth-oidc-client/blob/ea4e382b2607205297650523b96874a88e278bed/projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts#L94-L120

This causes the check session logic to malfunction because the session state check always takes place locally.
Are there any reasons why you don't invoke init() inside pollServerSessionRecur() function as it was in previous version?

And also it would be very convenient to be able to configure these variables:
https://github.com/damienbod/angular-auth-oidc-client/blob/ea4e382b2607205297650523b96874a88e278bed/projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts#L20-L21

valdian:
Refresh check session iframe regularly (always after 60 seconds), like it used to work before commit 1750ee252116a3bce4bd601b5d3fc4835a61dd99.

Fixes #756.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts
@@ -1,5 +1,5 @@
 import { Injectable, NgZone } from '@angular/core';
-import { BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, Observable, of } from 'rxjs';
 import { ConfigurationProvider } from '../config/config.provider';
 import { LoggerService } from '../logging/logger.service';
 import { EventTypes } from '../public-events/event-types';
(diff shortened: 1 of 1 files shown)


damienbod:
Hi @valdian Thanks for the PR. You will need to add some tests before we can merge. Can you help me a bit here, what does this PR fix or address? How can I reproduce the problem? I have tested the checkSession and this works for me (The way I test it, but I assume you use it different, so I probably missed smething)

Greetings and thanks for the PR Damien

damienbod:
PR is good, would like to merge this and add it back in. @FabianGosebrink @valdian How will we solve the tests?

FabianGosebrink:
THank you @valdian . Thanks for your contribution! Highly appreciated.

damienbod:
@valdian thanks for this, we will test and release in the next few days. Version 11.1.3

PR #780, opened 2020-06-05
remove id_token expire check

Zetanova:
The `id_token? should only be checked on receive to the client,
but not after it is persisted.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/authState/auth-state.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/authState/auth-state.service.spec.ts
@@ -180,7 +180,7 @@ describe('Auth State Service', () => {
             spyOn(authStateService as any, 'hasIdTokenExpired').and.returnValue(true);
             spyOn(authStateService as any, 'hasAccessTokenExpiredIfExpiryExists').and.returnValue(false);
             const result = authStateService.areAuthStorageTokensValid();
-            expect(result).toBeFalse();
+            expect(result).toBeTrue();
         });
 
         it('isAuthorized is true  and access_token is expired returns true', () => {
--- a/projects/angular-auth-oidc-client/src/lib/authState/auth-state.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/authState/auth-state.service.ts
@@ -26,7 +26,7 @@ export class AuthStateService {
         private publicEventsService: PublicEventsService,
         private configurationProvider: ConfigurationProvider,
         private tokenValidationService: TokenValidationService
-    ) {}
+    ) { }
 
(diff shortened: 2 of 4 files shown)


damienbod:
Hi @Zetanova @jamesikanos

Doing a lot of thinking about this. I see don't see a reason to allow an invalid id_token. If the validation for this is disabled, then the code will get complicated for some logout flows and by init and also allow a possible copy paste problem from old id_tokens which have expired.

I think that you can increase the lifespan of the id_token on the server, if you require this. If the token lifespan is short, this is usually for a reason. You can refresh you session if either the access token expires or the id_token expires.

So if the lifespan of the id_token is not important for you, then just increase this in the STS and all is good.

hope this is ok for you, and thanks for the PR.

Greetings Damien

Zetanova:
I made some research and come to the conclusion that the id_token 
should only be validated once on receive by the client. 
That is why the id_token expire time is always short (very short) 

The Authorization Code Lifetime and Identity Token Lifetime should be short
so that the client can validate the freshness of the authentication. 


robotdan made a list:
https://github.com/FusionAuth/fusionauth-issues/issues/110#issuecomment-481526544

jcfore21:
@Zetanova @damienbod 
Yeah I've been looking into the same thing for days.

I was thinking that the id_token only matters when initially requesting an access token.  From then on the only expiration that should matter is on the access token itself.  The reason the id_token has an expiration is so you can get back an access token and store it somehow (local storage in this case) within that time period.  From then on you should only be refreshing the access_token using a refresh token. OR just log the user out.   

Take a look at the diagram on [here](https://auth0.com/docs/flows/concepts/auth-code-pkce):

image

When looking at the code it seems we are always refreshing id_tokens and access tokens.  I'm not sure what the purpose of that is for the id_token.  Could you point us to an explanation why the code is doing that? @damienbod 

Thanks

damienbod:
@jcfore21 

The id_token is for your application, the access token is for your APIs. The id_token exp is required for the validation and for the logout and anytime the application refreshes the session or reads the claims, so not just for the first authn response. If the id_token has expired, then you no longer have a valid session and you should no longer be in an active client session.  If using refresh tokens as the link points too above, the id_token should also be validated. 

So the id_token should be only used when it is valid, which is not only by the first authn request. This is why the UI client checks if it's tokens are valid. The iat claim is used to validate that the id_token has been created close to the actual time, not the exp or the freshness of the authn.

Zetanova:
I am closing the PR.

There is one point to mention.
The current implementation checks every ~3sec for the expiration of  the id or access token,
but the expiration dates are well known and the cant expire premature 
and the timeout for the check can be extended to near the expiration date.

Issue #814, opened 2020-07-05
Angular 10 tslib warning


PartyArk:
With Angular 10 (and therefore tslib 2.0) I get this on `npm install `

`npm **WARN** angular-auth-oidc-client@11.1.3 requires a peer of tslib@^1.10.0 but none is installed. You must install peer dependencies yourself.`

I think the angular team recommends a new way of depending on tslib for libraries. Doesn't seem to make any difference to the functionality of the code.

damienbod:
hi @PartyArk Thanks for reporting.

@FabianGosebrink any ideas?

FabianGosebrink:
Yeah, we'll check that out and tackle that in the next days. Thanks for reporting!

PartyArk:
Thanks I think this is the relevant doc-  https://angular.io/guide/migration-update-libraries-tslib

FabianGosebrink:
IMHO as the PR is merged, we can close this one. Feel free to re-open if you have any further errors.

PartyArk:
Hi, I think this has regressed for 11.2.1 - the package.json  reference to tslib has reverted back to an old version.

Thanks!

PR #817, opened 2020-07-06
updating tslib

FabianGosebrink:
#814


damienbod:
Angular 8 = tslib 1.10.0
Angular 9 = tslib 1.10.0
Angular 10 = tslib 2.0.0

don't know how to support 2.0.0 without breaking older versions.

PR #836, opened 2020-08-01
Support for cookie storage have been added.

arv9135:
By default, angular-auth-oidc-clieint uses session storage that causes redirection to identity server page every time when opening the page in new instances of the browser. Also, session get cleared when tab gets closed. To fix that issue, we can use cookie storage instead of session storage.

Support for cookie storage has been added to extend the lifetime of the token in the browser across the tabs/windows.


damienbod:
looks great. We need to fix the tests. @FabianGosebrink any idea how we could do this? browserless tests...

FabianGosebrink:
Hey @arv9135 , thanks for the PR. Could you use prettier as a formatter in your vscode and re-save all the files you touched? Thanks.

damienbod:
Hi @arv9135 This would be nice to merge. Any chance of fixes?

Greetings Damien

guyz92:
There are some complications when using cookie.
Each of the Cookies only support a maximum size of 4096 byte.
The maximum size per cookie is computed using the length of key and value.
The "_authnResult" generated by some of OIDC Server such as Microsoft Identity Platform may exceed 4096 byte.
Base on my understanding, "_authnResult" mainly include the data retrieved from "/token" endpoint of the OIDC Server with additional state information generated by this library. If the size exceeded 4096 byte, the browser will ignore the creation of that particular cookie into the cookie storage of the browser. This could result in breaking the library itself.

Therefore, the cookie storage implementation will suffer from some form of complication and may not be suitable for every use case.

arv9135:
Sorry, I missed your comments. I will fix the conflicts. @FabianGosebrink Sure, will use prettier and save them.

PR #876, opened 2020-10-18
Fabiangosebrink/adding schematics

FabianGosebrink:


projects/schematics/src/ng-add/files/auth-config.module.ts
```diff
@@ -0,0 +1,32 @@
...
+import { AuthModule, OidcConfigService } from 'angular-auth-oidc-client';
+
+export function configureAuth(oidcConfigService: OidcConfigService): () => Promise<any> {
+    return () =>
+        oidcConfigService.withConfig({
+            stsServer: 'please-enter-sts',
+            redirectUrl: window.location.origin,
+            postLogoutRedirectUri: window.location.origin,
+            clientId: 'please-enter-clientId',
+            scope: 'please-enter-scopes',
```
damienbod, line 11: Would a comment help here for the scope configuration?

// 'openid profile '  + your scopes, offline_access if you refresh

projects/schematics/src/ng-add/actions/add-dependencies.ts
```diff
@@ -0,0 +1,32 @@
+import { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { addPackageJsonDependency, NodeDependency, NodeDependencyType } from '@schematics/angular/utility/dependencies';
+
+const dependenciesToAdd = [
+    {
+        name: 'angular-auth-oidc-client',
+        version: '11.3.0',
```
damienbod, line 7: Do we have to update this with every version update? Just a note for me

projects/schematics/src/collection.json
```diff
@@ -0,0 +1,9 @@
+{
+    "$schema": "../../../node_modules/@angular-devkit/schematics/collection-schema.json",
+    "schematics": {
+        "ng-add": {
+            "description": "Add my library to the project.",
```
damienbod, line 5: Shouls the description be 

Add OIDC lib to project?

projects/schematics/src/ng-add/actions/add-dependencies.ts
FabianGosebrink: We only have to update this if the mechanism for bootstrapping the auth in our lib changes. but to be consistent I would also update it with every release. I plan to improve that in the next version where the current version is being queries from npm directly via http request.

projects/schematics/src/collection.json
FabianGosebrink: Absolutely.

PR #888, opened 2020-10-30
UTC time

### LINKED ISSUE #175: Id_Token is rejected because of timing issue when server hour is different then client hour
I have both my OIDC Server and Angular App using angular-auth-oidc-client instance hosted on a server which has local time set to Belgium (UTC +1).

My PC is on UTC +2 so I receive the **authorizedCallback Validation, iat rejected id_token was issued too far away from the current time** error because of that. 

When I change server's time to be the same with my PC time, it works fine.

We should somehow take that into consideration when measuring time passed since the token was issued, given that it is very possible that clients of the library will not have the same local time as the server where the application is hosted.

bonaparte89:
Make sure to always use utc time when validating token.

closes #175

- tested and validated on several QA environments and already live in production.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "bugs": {
         "url": "https://github.com/damienbod/angular-auth-oidc-client/issues"
     },
-    "version": "11.2.1",
+    "version": "11.2.2",
     "scripts": {
         "ng": "ng",
         "build": "npm run build-lib",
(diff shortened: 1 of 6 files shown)


damienbod:
@bonaparte89 thanks LGTM

@FabianGosebrink ?

FabianGosebrink:
LGTM as well.

FabianGosebrink (approved):

Issue #877, opened 2020-10-19
Silent renew outside of i-frame.


DieterDS1983:
**Describe the bug**
We sometimes get the following error: Error: "codeFlowCodeRequest incorrect state"
After investigation we noticed a request to the silent renew page outside of the i-frame. The browser is serving this request from the cache since all request parameters resembles (same code, ...) an old request to the silent renew. (mostly not the previous one but the one before)
This request causes the error and the error causes the removal of all data from the session storage.

This is what we see in the Network tab.
First there are 3 valid request from the i-frame. the calls the authorize and then there is a redirect to the silent-renew page.
The fourth silent-renew seems t come from nowhere...
![network log](https://user-images.githubusercontent.com/18525722/96460612-ad904280-1223-11eb-8f47-060b1322b724.PNG)

**To Reproduce**
We where unable to reproduce this. It just happens from time to time.

**Expected behavior**
We expected no call to silent renew. the time for the next silent renew was 2 min later.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Widows 10
 - Browser: chrome
 - Version: 86

DieterDS1983:
Update.

We where able to reproduce the problem.
It is related to the browsers history. We noticed that when the silent renew happens, an item is added to the browsers history.
When we stay on the same page without any actions, the browser history is filling up with a new item on every silent renew.

When we press back and the previous item in the history was added by a silent-renew, than it is requesting the previous silent-renew with an old code. => error
This will also set al values in session storage to null, so new request will get an unauthorized.

steps to reproduce:
- Login to the application
- Navigate to a page (you will see the browser history is updated)
- Wait for the silent renew to happen (This should also update the browser history)
- After the silent renew click "back" in the browser. (or do a window.history.back();)
- now you should have an "incorrect state" error in the console.

tested with version 11.1.2 and 11.2.0.

Also tested with the older 10.0.15 version. Here we have the same behavior, but session storage is not reset.
So the next requests will still work.

DieterDS1983:
We also have a possible workaround.

We changed our silent renew page to remove the iframe after the request.
The next silent-renew will than add a new iframe. This avoids the creation of history items.
````javascript
window.onload = function () {
      /* The parent window hosts the Angular application */
      var parent = window.parent;
      /* Send the id_token information to the oidc message handler */
      var event = new CustomEvent('oidc-silent-renew-message', { detail: window.location });
      parent.dispatchEvent(event);

[...]

Issue #790, opened 2020-06-12
Silent Renew - incorrect state while using back/forward button


agorny:
Silent Renew mechanism breaks when back/forward browser buttons are used. Navigating back loads silent-renew.html with old authorization code and breaks authorization flow.

I used a code-flow-auto-login example with Keycloak as my identity provider. For faster-reproducing steps, I set Access Token Lifespan to 1 min.  

1. Visit app main page
2. Log in
3. Wait on the main page (/home) for the first silent renew call. 
4. Go to /forbidden page
5. Wait until silent renew call happens
6. Click back button

Browser console produces the following error:
![image](https://user-images.githubusercontent.com/11831976/84486489-c436f380-ac9d-11ea-8eff-ffefbc0a1f53.png)

Expected behavior
When using back/forward buttons (or programmatic history navigation) silent-renew.html should not dispatch oidc-silent-renew-message with old *code*.

 - OS: Ubuntu 18.04
 - Chrome Version 83.0.4103.61 (Official Build) (64-bit), Firefox 77.0.1 (64-bit)

![image](https://user-images.githubusercontent.com/11831976/84485314-41f9ff80-ac9c-11ea-8087-881dc87bbf1f.png)

In the above screen, there are 3 calls to silent-renew.html. The first one happened on the /home route, second on /forbidden route and the third one happens immediately after clicking the back button. You can see on the screen that query param for that call has an old code value (from the /home first call). After that, isAuthorize state is lost and the user is no longer logged in. Doing the same thing using links works fine. 

I think that https://github.com/damienbod/angular-auth-oidc-client/issues/608 was the result of the same bug, but the conversation sidetracked a little bit and the main reason was not discovered.

maaso:
I think I'm hitting this problem as well, in an implicit flow in my case. I can only trigger the problem if 2 or more silent renews have passed without navigating; subsequently calling `location.back()` will return an `authorizedCallback incorrect nonce` error.

I've tried setting the `ignoreNonceAfterRefresh` to true, but that seems to only apply to the code flow? Result is the same..

agorny:
@FabianGosebrink @damienbod If you need more information from me about the issue just let me know. I am pretty sure this happens every time if you follow the described steps.

GabrielGil:
May this be related to #788.

JoellyR:
[...]

PR #895, opened 2020-11-11
Fixed issue with browser history on silent renew redirect to IS.

Expelz:
Fix for #790 and #877


damienbod:
@Expelz Thanks!
LGTM I will test this in the different browsers and merge, probably Friday

Greetings Damien

Issue #837, opened 2020-08-03
When session state becomes blank app doesn't receive any event


Expelz:
Hello,
In the recursive function `pollServerSessionRecur` doesn't emitted any event when session state becomes blank it only logs message to the console:
https://github.com/damienbod/angular-auth-oidc-client/blob/23eed52e85a6e838f381b8bc5c3345bf6a446861/projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts#L110-L112

After some investigation I found that in 10 version such notification event was included inside `pollServerSessionRecur` function: 
https://github.com/damienbod/angular-auth-oidc-client/blob/1df7d5ec07f89862bedeee9f3dfad27a1b82c8fb/projects/angular-auth-oidc-client/src/lib/services/oidc.security.check-session.ts#L108-L111

Are there any reasons why you decided to remove `check session changed` event notification from 11 version?

Use case where it's important:
If user has opened two tabs and does logout process in one tab, then in the second tab our App doesn't receive any event and `check session` process will simply log next message:
`OidcSecurityCheckSession pollServerSession session_state is blank`
And user will be still logged in because isAuthenticated$ observable doesn't emit updated value.

WolfspiritM:
I'm also having trouble getting isAuthenticated$ to update correctly based on check session using code flow.
What is the correct way to check if the user is still logged in once the "changed" event arrives? I can't see any relation between isAuthorized$ and the check session response. I can get checkSessionChanged$ to fire if I log out from a different tab and I only get the "session_state is blank" error once the token expires.

The only "workaround" I found was
``` ts
        this.oidcSecurityService.checkSessionChanged$.subscribe((x) => {
          this.oidcSecurityService.logoffLocal();
          this.oidcSecurityService.checkAuthIncludingServer().subscribe((x) => {
            console.log('Auth', x);
          });
        });
```

But that feels a bit hacky...is that really the way it should be done?

Also: Is there a way to poll for a session?  Is it a good idea to do checkAuthIncludingServer every minute? I don't think the checksession endpoint works when logged out, right?

valdian:
Hi @Expelz,

where do you store the tokens and other authentication data? Do you set `storage: localStorage` in the config? Or do you use your own implementation of AbstractSecurityStorage?

[...]

PR #896, opened 2020-11-11
Multiple tabs don't receive any event when session state becomes blank.

Expelz:
Add checkSessionChanged event emmiting when session state becomes blank to notify about it all already opened tabs.
Related issue #837.


valdian (commented):
Have you tested it? I would say you should do `this.checkSessionChangedInternal$.next(true);`

Expelz:
> Have you tested it? I would say you should do `this.checkSessionChangedInternal$.next(true);`

@valdian It's my mistake, sorry about it. 
I already have fixed and tested it.

Thanks!

PR #905, opened 2020-11-25
Do not clear session state when refreshing session with refresh tokens

### LINKED ISSUE #899: Session State is lost on Refresh Token
**Describe the bug**
When a code pair (AT and RT) is acquired via PKCE Code Flow, a Session State value is set. This is used for the Check Session iFrame.

When a "Refresh Token" flow is performed, the session is lost.

The "Check Session" flow now fails.

**To Reproduce**
1. Acquire an AT/RT key pair
2. Observe that "Check Session" is working as expected
3. Refresh the token
4. Observe that "Check Session" reports error "OidcSecurityCheckSession pollServerSession session_state is blank"

**Expected behavior**
Given that the Refresh Token has come from a Session originally, I would expect the Session State to function, regardless of if the Access Token was acquired from Code flow or Refresh Token flow.

Perhaps an option could exist "retainSessionOverRefresh = true" to persist the session state when acquiring a new token.

 - OS: Windows
 - Browser Chrome
 - Version 86

valdian:
Fixes #899.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/flows/flows.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/flows/flows.service.ts
@@ -324,7 +324,9 @@ export class FlowsService {
             .pipe(
                 switchMap((userData) => {
                     if (!!userData) {
-                        this.flowsDataService.setSessionState(callbackContext.authResult.session_state);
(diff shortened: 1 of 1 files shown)


damienbod:
thanks, will test ad release this fix, hopefully Friday

PR #913, opened 2020-12-04
Use window object safely by injecting DOCUMENT

GabrielGil:
Using the global object window on the server-side produces errors, and it is discouraged.

The correct approach is to use the defaultView of the DOCUMENT injecting it.

Some features such as storage and crypto may not be available on the server, so we need to check before using them.

It is fine to go this way since there is no storage in the backend, thus there is no authentication happening there.

References: https://github.com/angular/universal/blob/master/docs/gotchas.md#window-is-not-defined

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
@@ -278,7 +278,7 @@ export class UrlService {
             return this.createAuthorizeUrl(codeChallenge, silentRenewUrl, nonce, state, 'none');
         }
 
-        // this.loggerService.logWarning('authWellKnownEndpoints is undefined');
+        this.loggerService.logWarning('authWellKnownEndpoints is undefined');
         return null;
     }
 
--- a/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
@@ -1,5 +1,5 @@
 import { HttpParams } from '@angular/common/http';
-import { Inject, Injectable } from '@angular/core';
+import { Injectable } from '@angular/core';
 import { oneLineTrim } from 'common-tags';
 import { ConfigurationProvider } from '../../config/config.provider';
 import { FlowsDataService } from '../../flows/flows-data.service';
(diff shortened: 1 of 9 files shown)


projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -280,7 +278,7 @@ export class UrlService {
             return this.createAuthorizeUrl(codeChallenge, silentRenewUrl, nonce, state, 'none');
         }
 
-        this.loggerService.logWarning('authWellKnownEndpoints is undefined');
+        // this.loggerService.logWarning('authWellKnownEndpoints is undefined');
```
FabianGosebrink, line 281: Is this comment on purpose? Maybe a leftover? Can be put back in or removed completely?

GabrielGil: Yes, it is a leftover. I'm changing it now. Thanks for the heads up :)

GabrielGil: Yes, it is a leftover. I'm changing it now. Thanks for the heads up :)

```suggestion
        this.loggerService.logWarning('authWellKnownEndpoints is undefined');
```

FabianGosebrink: You're welcome. That is why we do reviews. but thank YOU for the PR! Highly appreciated!

FabianGosebrink:
First Of all: Thanks to @GabrielGil for this PR. This is very cool. @damienbod  LGTM. Merge?

FabianGosebrink (approved):

damienbod:
@GabrielGil @FabianGosebrink Thanks! I will do the manual tests any release ASAP

damienbod:
Testing test and there is a problem with the prob build

pushed the changes to this branch: use-Window-object-safely-by-injecting-DOCUMENT

and revert the main.

Greetings Damien

GabrielGil:
Hi @damienbod. What is the status of this? Do you have an estimation of when this'll be released? Thanks!

FabianGosebrink:
I will get onto this maybe today.

Issue #921, opened 2020-12-21
How to handle errors from SSO server?


VladislavOshkanov:
Hi! 
I need to handle errors from SSO server in my app, for example, if I got 404 on SSO url.
How can I handle this error in my app in order to prevent app from crash and show error message to user?
Thanks
![image](https://user-images.githubusercontent.com/15010388/102768187-56d5ef00-43b3-11eb-811c-1825b9c75590.png)

FabianGosebrink:
So my first try would be first: Wrapping the promise in the APP_INITILIZER and catch the errors, then redirect the user or show an alert or something, or in the `main.ts` catch the bootstrap fail of the app and do something there. Hope this helps.

mgamal-dev:
App crash when sso is down why?

FabianGosebrink:
@mgamal-dev have you read my previous answer?

mgamal-dev:
Yeah, but I have some component will be public and not secured it is also crashed 
When I add catch block it doesn't mean the app will show this page correctly

FabianGosebrink:
You can catch the error in a promise with a reject/resolve. You have to wrap it again. Did not try it out but I think this should work.

PR #929, opened 2021-01-04
Added event to throw when config could not be loaded

FabianGosebrink:
#921


damienbod:
LGTM

will do some manual tests and merge and release 11.4.1

damienbod (approved):

Issue #904, opened 2020-11-24
[Bug Report] - "aud do not match" error


psavva:
**Describe the bug**
I have setup Keycloak and implemented the oidc client as per your example projects.
The issue is that when validating the new auth token with the refresh token flow, I recieve the following error:

As can be seen by the debug trace: "aud do not match: ehealth4u-ui,fhir ehealth4u-ui,fhir"
However, these two are the exact same: 
decodedIdToken.aud =>  ehealth4u-ui,fhir
newIdToken.aud        =>  ehealth4u-ui,fhir

The implementation of the validation is here:
https://github.com/damienbod/angular-auth-oidc-client/blob/db2ba2a81196f697d5e7c7ecf9e6908d86c4e35c/projects/angular-auth-oidc-client/src/lib/validation/state-validation.service.ts#L65

It seems the validtion isn't working as expected, as these two are equals, and should never have failed validation.

```
if (decodedIdToken.aud !== newIdToken.aud) {
            this.loggerService.logDebug(`aud do not match: ${decodedIdToken.aud} ${newIdToken.aud}`);
            return false;
}
```

```
**aud do not match: ehealth4u-ui,fhir ehealth4u-ui,fhir**
logger.service.ts:30 authorizedCallback pre, post id_token claims do not match in refresh
logWarning @ logger.service.ts:30
validateState @ state-validation.service.ts:205
getValidatedStateResult @ state-validation.service.ts:28
callbackStateValidation @ flows.service.ts:291
(anonymous) @ flows.service.ts:76
_next @ switchMap.ts:121
next @ Subscriber.ts:99
notifyNext @ switchMap.ts:166
_next @ innerSubscribe.ts:30
next @ Subscriber.ts:99
_next @ Subscriber.ts:139
next @ Subscriber.ts:99
.
.
.
(anonymous) @ zone.js:891
ZoneDelegate.invokeTask @ zone.js:421
onInvokeTask @ core.js:27492
ZoneDelegate.invokeTask @ zone.js:420
Zone.runTask @ zone.js:188
drainMicroTaskQueue @ zone.js:601
Promise.then (async)
scheduleMicroTask @ zone.js:584
ZoneDelegate.scheduleTask @ zone.js:410
Zone.scheduleTask @ zone.js:231
Zone.scheduleMicroTask @ zone.js:251
scheduleResolveOrReject @ zone.js:881
ZoneAwarePromise.then @ zone.js:1027
[...]

PR #950, opened 2021-01-30
Support aud arrays which are not ordered in id_token validation of refresh token process #904

FabianGosebrink:
Fixing #904 adding validation also for strings and arrays

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/utils/equality/equality.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/utils/equality/equality.service.ts
@@ -2,13 +2,40 @@ import { Injectable } from '@angular/core';
 
 @Injectable()
 export class EqualityService {
+    areRefreshEqual(value1: string | any[], value2: string | any[]) {
+        if (this.isNullorUndefined(value1)) {
+            return false;
+        }
+
+        if (this.isNullorUndefined(value2)) {
+            return false;
+        }
+
+        if (this.oneValueIsStringAndTheOtherIsArray(value1, value2)) {
+            console.log('oneValueIsStringAndTheOtherIsArray');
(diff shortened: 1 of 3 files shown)


projects/angular-auth-oidc-client/src/lib/utils/equality/equality.service.ts
```diff
@@ -2,13 +2,40 @@ import { Injectable } from '@angular/core';
 
 @Injectable()
 export class EqualityService {
+    areRefreshEqual(value1: string | any[], value2: string | any[]) {
```
damienbod, line 5: rename to isNewAudInIdTokenAfterRefreshTokenRequestValid

projects/angular-auth-oidc-client/src/lib/validation/state-validation.service.ts
```diff
@@ -62,8 +64,8 @@ export class StateValidationService {
         }
 
         // its aud Claim Value MUST be the same as in the ID Token issued when the original authentication occurred,
-        if (decodedIdToken.aud !== newIdToken.aud) {
-            this.loggerService.logDebug(`aud do not match: ${decodedIdToken.aud} ${newIdToken.aud}`);
+        if (!this.equalityService.areRefreshEqual(decodedIdToken?.aud, newIdToken?.aud)) {
+            this.loggerService.logDebug(`aud do not match: '${decodedIdToken?.aud}' '${newIdToken.aud}'`);
```
damienbod, line 68: change 

this.loggerService.logDebug(`aud do not match: '${decodedIdToken?.aud}' 

to

this.loggerService.logDebug(`aud in new id_token is not valid: '${decodedIdToken?.aud}'

PR #958, opened 2021-02-05
Added Title to Silent Renew IFrame

JerrettDavis:
According to WCAG and Deque, frames should have a title for accessibility reasons. I've added the title property to the generated IFrame to be compliant with accessibility standards.

References:
[Deque](https://dequeuniversity.com/rules/axe/4.0/frame-title?application=axeAPI)
[W3](https://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140408/H64)


damienbod:
Thanks @JerrettDavis 

LGTM

Greeting Damien

FabianGosebrink (approved):

Issue #961, opened 2021-02-08
ApplicationRef.isStable is always false when using this package


BillyBlaze:
**Describe the bug**
When you kick-off the setup (with [checkAuth](https://github.com/damienbod/angular-auth-oidc-client/blob/main/docs/public-api.md#checkauth-observable)) then this package will create intervals and timeouts. When using Service workers it's important to get the application into stable mode or else the new version isn't downloaded. However this is not possible because this package is making it unstable and it will never turn stable.
  
  
**To Reproduce**
1. Add this snippet somewhere (e.g. constructor of app.module.ts):
   ```ts
    this.applicationRef.isStable.pipe(
    	tap(isStable => console.log(`ApplicationRef::stable`, isStable))
    ).subscribe();
    ``` 
1. Initialize this package with calling checkAuth (we do it in the initializer, so that the guards are fully working with this package)
1. Notice that in the console the message 'ApplicationRef::stable false' is recieved but you will never get 'ApplicationRef::stable true'
    - When you do have a "stable" version you should see 'ApplicationRef::stable false' and than followed up by 'ApplicationRef::stable true'
  
  
**Expected behavior**
intervals and timeouts should be run outside the NgZone so that it wont interfere with the "stability" of Angular.

This package reaches stable for me when these two culprits are run outside angular:
https://github.com/damienbod/angular-auth-oidc-client/blob/710af82194bc9372d8a5ffedae00d94fc6f5417c/projects/angular-auth-oidc-client/src/lib/callback/intervall.service.ts#L18
https://github.com/damienbod/angular-auth-oidc-client/blob/03aa2ad53a5c4622300e7e283e2f4bd8535e1019/projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts#L132
  
  
**Potential fixes**
intervall.service.ts:
```ts
startPeriodicTokenCheck(repeatAfterSeconds) {
	const millisecondsDelayBetweenTokenCheck = repeatAfterSeconds * 1000;
	const update$ = new BehaviorSubject({});

	this.zone.runOutsideAngular(() => setInterval(() => this.zone.run(() => update$.next({})), millisecondsDelayBetweenTokenCheck))

	return update$;
}
```

check-session.service.ts:
```ts
this.zone.runOutsideAngular(() => {
	this.scheduledHeartBeatRunning = setTimeout(() => this.zone.run(pollServerSessionRecur), this.heartBeatInterval);
});
```

[...]

PR #965, opened 2021-02-10
ApplicationRef.isStable is always false when using this package

BillyBlaze:
**Describe the bug**
When you kick-off the setup (with [checkAuth](https://github.com/damienbod/angular-auth-oidc-client/blob/main/docs/public-api.md#checkauth-observable)) then this package will create intervals and timeouts. When using Service workers it's important to get the application into stable mode or else the new version isn't downloaded. However this is not possible because this package is making it unstable and it will never turn stable.
  
  
**To Reproduce**
1. Add this snippet somewhere (e.g. constructor of app.module.ts):
   ```ts
    this.applicationRef.isStable.pipe(
    	tap(isStable => console.log(`ApplicationRef::stable`, isStable))
    ).subscribe();
    ``` 
1. Initialize this package with calling checkAuth (we do it in the initializer, so that the guards are fully working with this package)
1. Notice that in the console the message 'ApplicationRef::stable false' is recieved but you will never get 'ApplicationRef::stable true'
    - When you do have a "stable" version you should see 'ApplicationRef::stable false' and than followed up by 'ApplicationRef::stable true'
  
  
**Expected behavior**
intervals and timeouts should be run outside the NgZone so that it wont interfere with the "stability" of Angular.

This package reaches stable for me when these two culprits are run outside angular:
https://github.com/damienbod/angular-auth-oidc-client/blob/710af82194bc9372d8a5ffedae00d94fc6f5417c/projects/angular-auth-oidc-client/src/lib/callback/intervall.service.ts#L18
https://github.com/damienbod/angular-auth-oidc-client/blob/03aa2ad53a5c4622300e7e283e2f4bd8535e1019/projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts#L132

#961

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/callback/intervall.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/callback/intervall.service.ts
@@ -1,5 +1,5 @@
 import { Injectable, NgZone } from '@angular/core';
-import { Subject, Subscription } from 'rxjs';
+import { Observable, Subscription } from 'rxjs';
 
 @Injectable({ providedIn: 'root' })
 export class IntervallService {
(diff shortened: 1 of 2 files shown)


damienbod:
@BillyBlaze Thanks for the PR. I'll go through this in the next few days, merge and release, up to my eyes in work.

Greetings Damien

damienbod:
@BillyBlaze Thanks works great. will be releasing in the 11.5.1 version

Greetings Damien

PR #995, opened 2021-03-01
Fabiangosebrink/adding get config method

FabianGosebrink:


projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts
```diff
@@ -72,14 +75,14 @@ export class RefreshSessionService {
       return of(null);
     }
 
-    const authWellknownEndpointAdress = this.configurationProvider.openIDConfiguration?.authWellknownEndpoint;
+    const { authWellknownEndpoint } = this.configurationProvider.getOpenIDConfiguration();
```
FabianGosebrink, line 78: add null check here

PR #1010, opened 2021-03-12
Add support custom params during token exchange

dopsonbr:
I have a use case where I need to pass custom params, like `roles`, during the token exchange. Are you open to accepting a PR that enables this behavior?

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
+++ b/projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
@@ -31,6 +31,7 @@ export interface OpenIdConfiguration {
   disableIatOffsetValidation?: boolean;
   storage?: any;
   customParams?: { [key: string]: string | number | boolean };
+  customTokenParams?: { [key: string]: string | number | boolean };
   eagerLoadAuthWellKnownEndpoints?: boolean;
 
   // Azure B2C have implemented this incorrectly. Add support for to disable this until fixed.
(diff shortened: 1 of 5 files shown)


damienbod:
@dopsonbr  LGTM thanks

I am planning to release today version 11.6.3

Will test and addd test.

Greetings Damien

dopsonbr:
My company uses an on-prem installation of [PING](https://www.pingidentity.com/en/software/pingfederate.html) so the STS isn't public.

damienbod:
@dopsonbr Looking into to this. Main problem is that the custom params are sent then for both the authorize and the token request with this change. I'm not sure other STS will work when send additional custom params to the token end point. I need to validate this first.

dopsonbr:
> @dopsonbr Looking into to this. Main problem is that the custom params are sent then for both the authorize and the token request with this change. I'm not sure other STS will work when send additional custom params to the token end point. I need to validate this first.

@damienbod I agree this is potentially problematic. What if I moved it into a new config parameter `customTokenParams` in the `OpenIdConfiguration` object?

damienbod:
@dopsonbr Yes this would be great, then the PR will not effect anything else.

Greetings Damien

dopsonbr:
@damienbod I just made the change and added tests. Please let me know what you think

damienbod:
Thanks @dopsonbr , works great. Will be released in version 11.6.4

Issue #985, opened 2021-02-25
Silent renew with refersh tokens broken after first failed request - Handle no connection use case


dome4:
**Describe the bug**
The silent renew check seems to be broken after the first request fails e.g. because of a connection problem. The result is  `OidcService code request` to my endpoint with an HTTPErrorResponse. I'm using refresh tokens and just turned the internet connection off. Is someone experiencing the same issue?

**To Reproduce**
1. Log the user in
2. Turn off internet connection in the browser
3. See error

**Expected behavior**
HTTP Error response should be handeled properly. If a *token not valid* response is emitted by the identity server, the user should be logged out. Otherwise,  the request should be retried in a specific intervall.

damienbod:
Hi @dome4 I tested this and it works, Tried with the example by disconnecting the internet and a second test revoking the refresh token. Both works. The authenticated is set to false which you can subscribe to.

Greetings Damien

dome4:
Hey @damienbod, sorry for the late reply and thanks for looking into this. I know it could be in the app's scope and probably not in the auth library, but for a PWA a user should not be set to isAuthenticated=false immediately if a token refresh fails. One reason for this could be the lack of internet connection. I'm thinking about distinguishing between two scenarios:

(1) an *invalid token response* from the auth server (often status code 400) -> isAuthenticated should be set to false and the user should be logged out immediately.
(2) other responses e.g. status code 0 (no internet connection) -> token refresh failed. The app cannot sync with the backend, but the user should be able to use the app as long as they had a valid access token at some point in the past.

Correct me if I'm on the absoultely wrong track, but it might be interesting for other PWA apps.



Translated with www.DeepL.com/Translator (free version)

damienbod:
I would makes sense to handle the no internet connection, thanks for the issue.

Greetings Damien

PR #1040, opened 2021-04-06
Silent renew with refresh tokens - handle no connection use case

dome4:
As discussed in #985, the periodical silent renew breaks after the first failed request. Here, the *no connection* use case should be handeled properly. The *no connection* check is done according to the following instructions:
- Angular - Getting error details (https://angular.io/guide/http#getting-error-details):
- use ProgressEvent instead of ErrorEvent: https://github.com/angular/angular/issues/34748

TODO: implement retry logic
- [x] RefreshTokenCallbackHandlerService
- [x] CodeFlowCallbackHandlerService
- [x] tests

Looking forward to opinions :)


FabianGosebrink:
Looks good to me so far. Can you add tests and the open checkboxes? Let us know if you need help. Thanks, you rock!

dome4:
@FabianGosebrink thanks for the feedback. I am currently trying to implement a fully offline capable pwa (https://github.com/dome4/angular-auth-oidc-client/commits/keycloak-pwa). So besides the retry logic, it would be helpful to be able to skip the service worker, which causes some problems in production. It would be possible to skip the auth-related requests via datagroups (https://angular.io/guide/service-worker-config#datagroups) and maxAge 0, but this comes with massive performance problems. Interceptors are also not possible because there's no way to acces the HttpClient used in the library from the app scope. Therefore, it would make sense to have the service worker skip the auth-related requests. Are there any concerns about this feature? 

I am aware that I am somewhat mixing up the original meaning of the pull requests. But the feature is necessary for me to realize the offline capability.

In addition I refactored the return of the mocked ConfigurationProdivder because otherwise the tests are broken:

```
const { stsServer } = this.configurationProvider.getOpenIDConfiguration();
```

Currently, the mocked ConfigurationProdivder returns null. Trying to destruct null results in an error. That might also be the reason why the pipeline for main failed.

dome4:
I added a simple retry logic to the requests executed in the following methods:
- AuthWellKnownDataService -getWellKnownDocument()
- SigninKeyDataService - getSigningKeys()
- LogoffRevocationService - revokeRefreshToken() & revokeAccessToken()
- UserService - getIdentityUserData()
- ParService - postParRequest()
In my opinion, these requests don't need complex retry logic but should be executed three times in total just because connection problems might occur in production.

Furthermore,  I added the same custom retry logic to the CodeFlowCallbackHandlerService like in RefreshTokenCallbackHandlerService. The tests are passing.

I'm looking forward to opinions and will start with the tests if you agree with the changes :)

projects/angular-auth-oidc-client/src/lib/api/data.service.ts
```diff
@@ -1,24 +1,37 @@
...
     });
   }
 
   post<T>(url: string, body: any, headersParams?: HttpHeaders) {
     const headers = headersParams || this.prepareHeaders();
+    let params = new HttpParams();
+
+    const { ngswBypass } = this.configurationProvider.getOpenIDConfiguration();
+    if (ngswBypass) {
+      params = params.set('ngsw-bypass', '');
```
FabianGosebrink, line 31: If this is a custom header, we have to prefix it with an `x-` so the header would be `x-ngsw-bypass`

dome4: Defining a constant for the header / param sounds good to me :+1: 

The first try of this feature was using the custom header `ngsw-byass`. The problem is that prefixing the header with `x-` will not work because of the following check in the service worker nsgw-worker.js:

```
onFetch(event) {
            ...
            if (req.headers.has('ngsw-bypass') || /[?&]ngsw-bypass(?:[=&]|$)/i.test(requestUrlObj.search)) {
                return;
            }
            ...
```
So the header must be exactly `ngsw-bypass`. Otherwise it will be ignored. Furthermore, I experienced the error that Keycloak rejects requests containing the custom header due to a CORS error because the custom header is not in the list of the allowed headers. I tried to add it somehow to keycloak but that was not easily possible and might also cause conflicts with other identity servers. 

Besides using a custom header, it's possible to use a **custom param** , which is what I would suggest. When using this approach a param will be appended to all auth-related requests: `https://identity-server.com/example?ngsw-bypass=`. and the CORS issues are gone.

I see that the description in the configuration.md is misleading and needs to be adjusted to the new approach. So my suggestion would be to define the constant `NGSW_CUSTOM_PARAM = ngsw-bypass` and update the description from `Adds the ngsw-bypass header to all requests` to `Adds the ngsw-bypass param to all requests`

FabianGosebrink: Sounds good to me. Please go with that. @damienbod ?

dome4:
I created the tests for the new functions and added a test help to test the retry logic. Since I couldn't find any other helper, I created the file under `projects/angular-auth-oidc-client/test/create-retriable-stream.helper.ts`. There might be better paths so feel free to feedback and I'll update the location :)

damienbod:
@dome4 Thanks for the PR!   Thanks @FabianGosebrink for the reviewing.

LGTM 

I'll merge and test today. All good we could release then.

projects/angular-auth-oidc-client/src/lib/flows/callback-handling/refresh-token-callback-handler.service.spec.ts
```diff
@@ -45,6 +46,13 @@ describe('RefreshTokenCallbackHandlerService', () => {
   });
 
   describe('refreshTokensRequestTokens', () => {
+    const HTTP_ERROR = new HttpErrorResponse({});
+    const CONNECTION_ERROR = new HttpErrorResponse({
+      error: new ProgressEvent('error'),
+      status: 0,
+      statusText: 'Unknown Error',
+      url: 'https://identity-server.test/openid-connect/token',
+    });
```
FabianGosebrink, line 55: add new line after line 55

projects/angular-auth-oidc-client/src/lib/flows/callback-handling/code-flow-callback-handler.service.spec.ts
```diff
@@ -108,6 +109,13 @@ describe('CodeFlowCallbackHandlerService', () => {
   });
 
   describe('codeFlowCodeRequest ', () => {
+    const HTTP_ERROR = new HttpErrorResponse({});
+    const CONNECTION_ERROR = new HttpErrorResponse({
+      error: new ProgressEvent('error'),
+      status: 0,
+      statusText: 'Unknown Error',
+      url: 'https://identity-server.test/openid-connect/token',
+    });
```
FabianGosebrink, line 118: add new line after 118

projects/angular-auth-oidc-client/src/lib/api/data.service.ts
```diff
@@ -1,24 +1,38 @@
-import { HttpHeaders } from '@angular/common/http';
+import { HttpHeaders, HttpParams } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
+import { ConfigurationProvider } from '../config/config.provider';
 import { HttpBaseService } from './http-base.service';
 
+const NGSW_CUSTOM_PARAM = 'ngsw-bypass';
```
FabianGosebrink, line 7: add new line after 7

FabianGosebrink:
This looks really good to me, three really small new lines missing. Incredible job, really good. Thank you very much. Love the way you tackled down the problem and then documented and tested it in a beautiful manner. Kudos to you!

FabianGosebrink:
For me this is done, code looks good. @damienbod please test and if good then we can merge imho. @dome4 thanks for this amazing feature and nice code. Thank you.

PR #1047, opened 2021-04-14
Store singing keys as fallback

dome4:
I was thinking about how to cache the JWKS in `callbackHistoryAndResetJwtKeys()` of the `HistoryJwtKeysCallbackHandlerService` because the method is used in every refresh cycle. In production, the token refresh might succeed, but if the connection is lost before requesting the JWKS, the entire pipeline will break after two retries. However, storing the result like in `getAuthWellKnownEndPoints()` of the `AuthWellKnownService` is not feasible because of possible key rotations by the identity server. Therefore, I decided on a *network-first* strategy that tries the request for three times but falls back to the stored keys in case of failure. 

In addition, the retrieved JWKS always replace the stored keys, which minimizes the risk of validating the tokens against expired keys.

Any thoughts?

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/history-jwt-keys-callback-handler.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/history-jwt-keys-callback-handler.service.ts
@@ -13,6 +13,8 @@ import { FlowsDataService } from '../flows-data.service';
 import { ResetAuthDataService } from '../reset-auth-data.service';
 import { SigninKeyDataService } from '../signin-key-data.service';
 
+const JWT_KEYS = 'jwtKeys';
+
 @Injectable()
 export class HistoryJwtKeysCallbackHandlerService {
   constructor(
@@ -102,10 +104,10 @@ export class HistoryJwtKeysCallbackHandlerService {
   }
 
   private storeSigningKeys(jwtKeys: JwtKeys) {
-    this.storagePersistanceService.write('jwtKeys', jwtKeys);
+    this.storagePersistanceService.write(JWT_KEYS, jwtKeys);
   }
 
   private readSigningKeys() {
-    return this.storagePersistanceService.read('jwtKeys');
+    return this.storagePersistanceService.read(JWT_KEYS);
   }
 }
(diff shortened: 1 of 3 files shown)


damienbod:
Hi @dome4 thanks for the PR

Thinking out loud now. Need to think about this a bit. I think that this is cached anyway by the browser as well use a HTTP GET request. 

One problem would be with key rotation. If we read the keys first from storage , we might be using old keys and validation will fail. This would probably never happen because the session storage gets clean up once we close the tab, browser etc., so not really a problem. 

If the HTTP request fail this is awesome. 

So maybe it good. 

@FabianGosebrink  @dome4 Would love your thoughts on this

Greetings Damien

dome4:
I totally agree with the concerns about key rotation. That's the reason why I tried to implement a network-first strategy: the idea is to try the request for three times (`retry(2)` in [SigninKeyDataService ](https://github.com/damienbod/angular-auth-oidc-client/blob/5061f75cd5d200dc3137b74515a4325b37ac80bc/projects/angular-auth-oidc-client/src/lib/flows/signin-key-data.service.ts#L29)). If and only if the request fails for three times, the stored keys are used as a fallback (first `catchError()` block in [HistoryJwtKeysCallbackHandlerService](https://github.com/damienbod/angular-auth-oidc-client/blob/6cfa8a92e0360898e23c30df659defe868326794/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/history-jwt-keys-callback-handler.service.ts#L52)). Thus, in the most cases, the stored keys are skipped.

Furthermore, caching get requests through the browser is not suitable for me, as I am trying to implement a PWA that stores the keys in local storage to be able to launch the app event when offline. In this case the session could be already destroyed when restarting the app.

damienbod:
@dome4 thanks, I think this is good and we add the feature. I'll test this once the code is good. @FabianGosebrink can you do a code review as well, LGTM, needs some tests.

Greetings Damien

projects/angular-auth-oidc-client/src/lib/flows/callback-handling/history-jwt-keys-callback-handler.service.ts
```diff
@@ -88,4 +100,12 @@ export class HistoryJwtKeysCallbackHandlerService {
   private resetBrowserHistory() {
     window.history.replaceState({}, window.document.title, window.location.origin + window.location.pathname);
   }
+
+  private storeSigningKeys(jwtKeys: JwtKeys) {
+    this.storagePersistanceService.write('jwtKeys', jwtKeys);
```
FabianGosebrink, line 105: Make this a constant in the class to avoid having different keys on read/write

FabianGosebrink:
Code looks absolutely fine, just one small comment. coverage dropped a bit, would be nice to have this staying at least the same. Good Job.

PR #1051, opened 2021-04-16
fix: use navigateByUrl to fix url params encoding

### LINKED ISSUE #1028: Angular Router navigations should be (probably) triggered with navigateByUrl() function
**Describe the bug**
Currently all side effects or `catchError` functions use `this.router.navigate()` function to redirect to one of the predefined routes
This works great but fails when users enter URL with query params.
(i.e http://example.com?foo=5&bar=10)

This actually is a known bug in Angular Router
(see https://github.com/angular/angular/issues/28917)

The workaround would be to use `navigateByUrl()` instead.
(for example: `this.router.navigateByUrl(this.configurationProvider.openIDConfiguration.postLoginRoute);`)

**To Reproduce**
1. Go to 'http://example.com?foo=5&bar=10'

2. Set postLoginRoute to '/?foo=5&bar=10

3. See error in console:
**(Error: Uncaught (in promise): Error: Cannot match any routes. URL Segment: '%3Ffoo%3D5&bar%3D10')**

**Expected behavior**
After navigation there shouldn't be any error in console and the Address bar should contain URL with query params

**Screenshots**
N/A

 - OS: Win 10
 - Browser chrome
 - Version 89.0.4389.90 

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

TomaszKasowicz:
Change `router.navigate()` calls to `router.navigateByUrl()`

Fixes #1028

### DIFF
--- a/docs/guards.md
+++ b/docs/guards.md
@@ -19,7 +19,7 @@ export class AuthorizationGuard implements CanActivate {
         console.log('AuthorizationGuard, canActivate isAuthorized: ' + isAuthorized);
 
         if (!isAuthorized) {
-          this.router.navigate(['/unauthorized']);
+          this.router.navigateByUrl('/unauthorized');
           return false;
         }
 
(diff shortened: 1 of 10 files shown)


damienbod:
@TomaszKasowicz thank you! LGTM

@FabianGosebrink good for you as well?

FabianGosebrink:
Yeah is fine for me. We can merge imho. @TomaszKasowicz thank you very much!

Issue #1060, opened 2021-04-23
Use a different key than `redirect` to store redirect route when using autologin


PackBjamin:
We experienced anomalous behavior with our localStorage['redirect'] value being "randomly" cleared out. Further investigation found the culprit was the angular-auth-oidc-client code:

auto-login-service.js
```javascript
const STORAGE_KEY = 'redirect';
```
I'm not familiar with the code base. My initial scan only found this usage. But it might be a good idea to use something like a library wide prefix for localStorage, etc. keys? For example:

```javascript
const LIBRARY_KEY_PREFIX = '__autoLoginSvc__';
const STORAGE_KEY = LIBRARY_KEY_PREFIX + 'redirect';
```

inloox-dev:
Wir erleichtern Ihnen den Einstieg mit InLoox Planungsvorlagen und stellen die Offline-Funktion detailliert vor.
InLoox Fehler wurde zur Nachverfolgung erstellt
[https://gallery.mailchimp.com/6fb276e204c3e429e34db0e30/images/43782466-0ad5-4511-953f-282e94d647e5.jpg] <https://www.inloox.de/?link=9671>
Guten Tag,


Fehler 25295 wurde zur Nachverfolgung erstellt

Titel: [damienbod/angular-auth-oidc-client] auto-login-service.ts uses localStorage['redirect'] (#1060)

Herzliche Grüße
Dein InLoox Dev-Team

Fehler anzeigen<https://inloox.visualstudio.com//InLoox%20PM/_workitems/edit/25295>
FOLGEN SIE UNS!
[Twitter]<https://twitter.com/inloox_gmbh>
[Facebook]<https://www.facebook.com/InLoox>
[Newsletter-Anmeldung]<https://www.inloox.de/about-us/contact-us/?mode=newsletter>
[Google Plus]<https://plus.google.com/+InlooxDe/posts>
[YouTube]<https://www.youtube.com/user/InLoox>
[InLoox auf Instagram]<https://www.instagram.com/inlooxgmbh/>
Impressum<https://www.inloox.de/impressum/>                             Datenschutzbestimmungen<https://www.inloox.de/datenschutzbestimmungen/>
Copyright © 2019 InLoox GmbH, All rights reserved.

________________________________
From: PackBjamin ***@***.***>
Sent: Friday, April 23, 2021 8:16:07 PM
To: damienbod/angular-auth-oidc-client ***@***.***>
Cc: Subscribed ***@***.***>
[...]

PR #1067, opened 2021-04-25
Fix redirect getting overwritten in localStorage

JerrettDavis:
Issue #1060 refers to application's having their own internal `redirect` record in `localStorage` overwritten by the `AutoLoginService`. The `StoragePersistenceService` already handles prefixing stored values to remedy this issue. I've converted the `AutoLoginService` to use `StoragePersistenceService` instead of `localStorage`.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -74,6 +74,10 @@ export class CustomStorage implements AbstractSecurityStorage {
     public remove(key: string): void {
         ...
     }
+    
+    public clear(): void {
+        ...
+    }
 }
 ```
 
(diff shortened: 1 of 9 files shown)


FabianGosebrink:
You are on fire! Thanks :-) We will look into this, a bit short of time currently, and then merge this. Thanks for your help, we really appreciate it!

JerrettDavis:
We've been using the library for a bit now, so I'm always happy to lend a hand. On that note, I still need to fix the unit tests before this can be merged. I should have that done shortly.

JerrettDavis:
I've updated all the necessary unit tests to make them compatible with the change away from `localStorage`. One minor change I implemented throughout was the inclusion of a `clear` method on the `AbstractSecurityStorage` class. This shouldn't be at all breaking, but it enables `localStorage`-based Unit Tests to perform post-test cleanup, if necessary. As it stands now, it shouldn't be necessary as the mock used for testing doesn't actually persist anything, but previously it was and I didn't want any sort of feature-regression.

FabianGosebrink:
Glad to hear that you are using the lib :-) Thanks for being part of this then :) We have to test the AutoLogin Good and the Popup. @damienbod and I will run the tests and then this will land in V12. Thanks!

Issue #1072, opened 2021-04-29
Add return types to applicable targets in OidcSecurityService


JerrettDavis:
**Is your feature request related to a problem? Please describe.**
There are several methods, fields, and other constructs in `OidcSecurityService` that do not declare any return types (even `any` would be sufficient for many). To better represent the intentions of the class, I believe all applicable constructs should have return types added.

**Describe the solution you'd like**
Return types added to all constructs in `OidcSecurityService`

**Describe alternatives you've considered**
_None_

**Additional context**
_None_

Issue #1071, opened 2021-04-28
Adding JSDoc comments throughout


JerrettDavis:
**Is your feature request related to a problem? Please describe.**
It appears that while there's excellent documentation for the library, very little of the code itself has attached in-code documentation. This often requires me to look directly in the library's code to really understand what a method is doing. Implementing JSDocs throughout would allow IDE code-completion and code-hints to better help other developers.

**Describe the solution you'd like**
Add JSDocs to all publicly facing methods.

**Describe alternatives you've considered**
_None_

**Additional context**
_None_

FabianGosebrink:
Hey, thanks for the input. Yes, I would love to do this while we want to keep the documentation to the public `OidcSecurityService` as he provides a facade as entrypoint for the library.

I would like to see this asap, so we can branch from main and do a PR from there. Would you like to help with that?

Thanks

JerrettDavis:
I'm more than happy to start up a branch and get to documenting. I should be able to get on it here in a few hours after I wrap up my day's work.

FabianGosebrink:
Take it easy, no rush. I can definitely review and help you with that. Thanks!

PR #1073, opened 2021-04-29
Document public facing API

JerrettDavis:
Documented the publicly facing API.

#1071 

Can we tackle #1072 in here as well? Return types can be part of the docs as well, would like to add that in one go.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
+++ b/projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
@@ -26,6 +26,9 @@ export interface OpenIdConfiguration {
   /**
    * This is this scopes which are requested from the server from this client.
    * This must match the STS server configuration.
+   * The 'openid' scope is required. The 'offline_access' scope can be requested when using refresh tokens 
+   * but this is optional and some STS do not support this or recommend not requesting this even when using 
+   * refresh tokens in the browser.
    */
   scope?: string;
   /**
@@ -63,10 +66,11 @@ export interface OpenIdConfiguration {
    */
   usePushedAuthorisationRequests?: boolean;
   /**
-   * 	A token obtained by using a refresh token normally doesn't contain a nonce value.
-   * 	However, some OIDC endpoint implementations do send one. The library checks to see if the nonce is present.
-   * 	Note that if the nonce value is present, it will not be verified.
-   * 	Default value is false
+   * A token obtained by using a refresh token normally doesn't contain a nonce value.
+   * However, some OIDC endpoint implementations do send one. The library checks to see if the nonce is present.
+   * Note that if the nonce value is present, it will not be verified.
+   * This is not recommended, if the STS returns a nonce in the refresh.
+   * Default value is false
    */
   ignoreNonceAfterRefresh?: boolean;
   /**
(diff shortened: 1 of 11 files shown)


JerrettDavis:
I have documented the entirety of the public API and the configuration. I have also added return types to all applicable fields and methods.

FabianGosebrink (approved):

projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
```diff
@@ -1,43 +1,147 @@
...
   clientId?: string;
+  /**
+   * 'code', 'id_token token' or 'id_token' Name of the flow which can be configured.
+   * You must use the 'id_token token' flow, if you want to access an API or get user data from the server.
+   * The access_token is required for this, and only returned with this flow.
+   */
   responseType?: string;
+  /**
+   * This is this scopes which are requested from the server from this client.
+   * This must match the STS server configuration.
```
damienbod, line 28: Can you append this:

The openid scope is required. This offline_access scope can be requested when using refresh tokens but this is optional and some STS do not support this or recommend not requesting this even when using refresh tokens in the browser.

projects/angular-auth-oidc-client/src/lib/oidc.security.service.ts
```diff
@@ -58,96 +76,164 @@ export class OidcSecurityService {
...
+   * https://tools.ietf.org/html/rfc7009
+   *
+   * @param refreshToken The access token to revoke.
+   */
+  revokeRefreshToken(refreshToken?: any): Observable<any> {
     return this.logoffRevocationService.revokeRefreshToken(refreshToken);
   }
 
+  /**
+   * Creates the ENS session URL which can be used to implement an alternate server logout.
```
damienbod, line 235: ENS => END

projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
```diff
@@ -1,43 +1,147 @@
...
+  /**
+   * Activates Pushed Authorisation Requests for login and popup login.
+   * Not compatible with Iframe renew.
+   */
   usePushedAuthorisationRequests?: boolean;
+  /**
+   * 	A token obtained by using a refresh token normally doesn't contain a nonce value.
+   * 	However, some OIDC endpoint implementations do send one. The library checks to see if the nonce is present.
+   * 	Note that if the nonce value is present, it will not be verified.
+   * 	Default value is false
```
damienbod, line 69: append

This is not recommended, if the STS returns a nonce in the refresh.

damienbod:
@JerrettDavis LGTM, just 3 comments

Greetings Damien

projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
JerrettDavis: This is now corrected.

projects/angular-auth-oidc-client/src/lib/oidc.security.service.ts
JerrettDavis: This is now corrected.

projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
JerrettDavis: This is now corrected.

PR #1078, opened 2021-04-29
fix(randomService): fix misuse of Uint8Array

### LINKED ISSUE #1070: Uint8Array Error Firefox (old version)
Hi,

One requirement for our client is to use the app in an older version of Firefox (52.9). However, it seems that something breaks in your side. After investigation, I was able to find the culprit: Uint8Array

Assuming this test case:

```javascript
const case1 = new Uint8Array(2)
const case2 = new Uint8Array(5/2)
```
We have the following results:
**_In recent version of Chrome browser_**
```
case1 = Uint8Array(2) [0, 0]
case2 = Uint8Array(2) [0, 0]
```

**_In Firefox 52 (our requirement)_**
```
case1 = Uint8Array(2) [0, 0]
case2 = TypeError: invalid arguments
```

It seems **odd values** divided by 2 raises an Error in this old version of Firefox. Maybe in others, I didn't investigate.

Maybe replacing all odd values used in your library by even values (which does the same effect), for example this line:
https://github.com/damienbod/angular-auth-oidc-client/blob/b3b2621349f9ca1725e26d59744262ef6e005888/projects/angular-auth-oidc-client/src/lib/flows/flows-data.service.ts#L56

Or use Math.floor() to be sure to have an integer.
`new Uint8Array(Math.floor(length || length) / 2)`

https://github.com/damienbod/angular-auth-oidc-client/blob/b3b2621349f9ca1725e26d59744262ef6e005888/projects/angular-auth-oidc-client/src/lib/flows/random/random.service.ts#L20

Thank you in advance and have a nice day

kmathy:
Necessary for compatibility with old browsers. Tested with Firefox 52, has mentioned in the issue

fixes #1070

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/flows/random/random.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/flows/random/random.service.ts
@@ -17,7 +17,7 @@ export class RandomService {
     }
 
     const length = requiredLength - 6;
-    const arr = new Uint8Array((length || length) / 2);
+    const arr = new Uint8Array(Math.floor((length || length) / 2));
     if (this.getCrypto()) {
       this.getCrypto().getRandomValues(arr);
     }


damienbod:
LGTM @FabianGosebrink

FabianGosebrink (approved):

PR #1089, opened 2021-05-11
Bugfix: the "use" attr on the jwks key is optional if only one key is present

andreaslarssen:
I couldn't find any contribution docs, and wasn't able to set this up locally, so I coded this "blindly". If someone could verify my changes, that would be great. 

One remark: Is the check for multiple key candidates needed? Removing it would result in the client using whatever key was defined last, and either the verification will fail (wrong key), or it will be verified (right key, by chance). Would not uncover an error in the jwks endpoint, but is that really up to the client to uncover?

EDIT: Could this be released as a patch of version 11?

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/validation/token-validation.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/validation/token-validation.service.ts
@@ -307,36 +307,43 @@ export class TokenValidationService {
 
     let isValid = false;
 
-    if (!headerData.hasOwnProperty('kid')) {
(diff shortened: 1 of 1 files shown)


damienbod:
@andreaslarssen Thanks for the PR, I'll review and merge to 11 is possible

andreaslarssen:
@damienbod thanks. Let me know if there's anything wrong with it, or if you have other inputs, and I'll fix as soon as possible. Also let me know your thoughts on my comment on the validation of no of keys

damienbod:
thanks, released in version 11.6.9

PR #1117, opened 2021-06-07
Fabiangosebrink/improving the migration to v12 docs

FabianGosebrink:


docs/site/angular-auth-oidc-client/docs/migrations/v11-to-v12.md
```diff
@@ -281,137 +146,318 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
...
 ```
 
-#### New
+### AuthWellKnownEndpoints are now part of the config
 
-```typescript
-this.oidcSecurityService.checkAuth().subscribe(({ isAuthenticated }) => {
-  console.log('app authenticated', isAuthenticated);
-});
+In the previous version the `AuthWellKnownEndpoints` have been a special parameter which you could pass. The type is the same but the property is now part of the configuration. The parameter to pass is in the `withConfig` method was removed.
```
JerrettDavis, line 173: For clarity, I would rephrase this to be:

In the previous version, `AuthWellKnownEndpoints` was a separate parameter you could provide alongside the config in `withConfig` to configure your STS Server's Well-Known Endpoints. This parameter still exists, and it is of the same type, but it's now a parameter on the config object rather than being a separate parameter on `withConfig`

docs/site/angular-auth-oidc-client/docs/migrations/v11-to-v12.md
```diff
@@ -281,137 +146,318 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
...
+  customParamsCodeRequest?: { [key: string]: string | number | boolean };
+  // ...
+}
+```
+
+## Service `OidcSecurityService`
+
+### Getter for `configuration` and object `PublicConfiguration` replaced by method
+
+The getter for the active configuration was removed as well as the object `PublicConfiguration`. As the `AuthWellKnownEndpoints` are now part of the config the property was replaced with methods to get the configuration directly. Please read below how to get the current used configuration.
```
JerrettDavis, line 244: Minor grammatical mistakes make this a bit confusing. I would correct like so:

The getter for the active configuration was removed as well as the object `PublicConfiguration`. Since the `AuthWellKnownEndpoints` parameter is now part of the config, the property was replaced with methods to get the configuration directly. Please read below **on** how to get the current**ly** used configuration.

docs/site/angular-auth-oidc-client/docs/migrations/v11-to-v12.md
```diff
@@ -281,137 +146,318 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
...
+
+New
+
+```ts
+const accessToken = this.oidcSecurityService.getAccessToken();
+```
+
+### `checkAuth()` returning `LoginResponse` instead of boolean
+
+The `checkAuth()` method is now not only returning if you are authenticated or not but instead a complete object containing:
```
JerrettDavis, line 293: Reword for clarity

The `checkAuth()` method no longer returns a boolean value denoting the authentication status. Instead, an object is returned containing:

docs/site/angular-auth-oidc-client/docs/migrations/v11-to-v12.md
```diff
@@ -281,137 +146,318 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
...
+
+New:
+
+```ts
+this.oidcSecurityService.checkAuth().subscribe(({ isAuthenticated }) => {});
+```
+
+### `authorize(...)` has new `configId` as first parameter
+
+Because multiple configs have been introduce with V12, now the `configId` is the first parameter.
```
JerrettDavis, line 352: Minor rewording for clarity:

Because V12 introduces multiple configs, the first parameter is now `configId`.

docs/site/angular-auth-oidc-client/docs/migrations/v11-to-v12.md
```diff
@@ -281,137 +146,318 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
...
+
+```ts
+const authOptions = {...}
+const configIdOrNull = ./*...*/;
+this.oidcSecurityService.authorize(configIdOrNull, authOptions)
+```
+
+### `logoffAndRevokeTokens(...)` has new `configId` as first parameter
+
+Because multiple configs have been introduce with V12, now the `configId` is the first parameter.
```
JerrettDavis, line 371: Minor rewording for clarity:

Because V12 introduces multiple configs, the first parameter is now `configId`.

docs/site/angular-auth-oidc-client/docs/migrations/v11-to-v12.md
```diff
@@ -281,137 +146,318 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
...
+
+```ts
+const authOptions = {...}
+const configIdOrNull = ./*...*/;
+this.oidcSecurityService.logoffAndRevokeTokens(configIdOrNull, authOptions)
+```
+
+### `logoff(...)` has new parameters
+
+Because multiple configs have been introduce with V12, now the `configId` is the first parameter and `AuthOptions` as a second.
```
JerrettDavis, line 390: Minor rewording for clarity:

Because V12 introduces multiple configs, the first parameter is now `configId` and the second parameter is `AuthOptions`.

docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
```diff
@@ -5,29 +5,30 @@ sidebar_position: 2
...
-You can pass three properties into the `AuthModule.forRoot(...)` method to configure the lib.
-
-```
-AuthModule.forRoot({
-  authWellKnown?: AuthWellKnownEndpoints;
-  config?: OpenIdConfiguration;
-  loader?: Provider;
-})
-```
+The lib has to be configured with the values of your choice. You can to that either with a static config or loading the config from an HTTP endpoint and map the values to the format the library expects.
```
JerrettDavis, line 8: I would rephrase for clarity, like so:

Prior to using the library, you must configure it with the appropriate values for your environment. You can either configure the application statically, by providing the configuration values at design-time, or you can fetch the configuration from an HTTP endpoint. Configurations loaded from an HTTP endpoint must be mapped to the format the library expects.

docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
```diff
@@ -65,60 +66,284 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
...
-      authWellKnown: {
-        /* YOUR AUTHWELLKNOWN VALUES HERE */
-      },
-    }),
-```
+- Type: string
+- Required: false
 
-## Config Values
+To identify a configuration a new parameter called `configId` was introduced. This is being filled from the lib when the user does not set it explicitly in the config. If set, the configured value is taken. It is optional. For a single config this param is not really important. For multiple configs this param is important.
```
JerrettDavis, line 76: I would rephrase for clarity, like so:

To identify a configuration a new parameter called `configId` was introduced. If you do not explicitly set this value, the library will generate and assign the value for you. If set, the configured value is used. The value is optional, and for single-configuration projects, `configId` is largely unimportant. For multiple-configuration projects, it is important to set a unique `configId` for each separate configuration.

(Note: If it's required to assign a `configId` for multi-config projects, I would also add that to the above verbiage)

JerrettDavis (commented):
There are several pieces that could be rephrased for clarity. I've included my suggestions in-line.

Issue #1113, opened 2021-06-06
Add configuration to disable or enable id_token expired check


damienbod:

damienbod:
fixed, will be released in V12

Issue #825, opened 2020-07-15
Question about the validateState and checks for expiration


gjberghuis:
We like to use the lifetime of our accessToken as leading to determine if we need to refresh them or not. 

I have two questions:

1. The last check in the validateState of state-validation.service.ts checks if the id token is not expired:

```
 if (!this.tokenValidationService.validateIdTokenExpNotExpired(toReturn.decodedIdToken)) {            
    this.loggerService.logWarning('authorizedCallback id token expired');
    toReturn.state = ValidationResult.TokenExpired;
    this.handleUnsuccessfulValidation();
    return toReturn;
 }
```

We would like to be able to skip this check. And in addition to that we would like to be able to check of the accessToken is not expired. This also happens in the callback.services.ts:

`         const accessTokenHasExpired = this.authStateService.hasAccessTokenExpiredIfExpiryExists();`

2. In the callback.services.ts we also want to decide if we want to check the accessToken lifetime, idToken lifetime or both of them:

```
     const idTokenHasExpired = this.authStateService.hasIdTokenExpired();
     const accessTokenHasExpired = this.authStateService.hasAccessTokenExpiredIfExpiryExists();

     if (!idTokenHasExpired && !accessTokenHasExpired) {
        return of(null);
     }
```

I hope you can help us with this or maybe we are lacking some knowledge about the importance of the identityToken lifetime, in that case, please help us with some explanation.

damienbod:
I consider disabling the id_token lifetime check per configuration. I have had this request quite often.

Thanks for reporting Damien

andreaslarssen:
Also experiencing this. From what I can see, there's nothing in the RFC that indicates that the TTL of the id_token should be an indicator of how long the session of an external app should last.

damienbod:
we add support for this in V12

damienbod:
added in V12, default is with the id_token check, you can config this to stop the check.

```
enableIdTokenExpiredValidationInRenew: false
```

Thanks for reporting

PR #1126, opened 2021-06-11
Disable id token renew expired check in renew process

damienbod:
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/1113
fixes https://github.com/damienbod/angular-auth-oidc-client/issues/825

enableIdTokenExpiredValidationInRenew can now be used to disable this check. default value is true


FabianGosebrink:
LGTM

Issue #1107, opened 2021-06-03
Silent Renew process doesn't always start with a protected route


JerrettDavis:
**Describe the bug**
When using `AutoLoginGuard` globally, rather than manually using `CheckAuth` or similar, the silent renew process fails to start until after the user has been reauthenticated at least one time. This leads to the user being redirected to the login page when navigating between secured routes, even after they've previously been authenticated. 

Once the user has been logged in a second time, the silent renew process usually behaves as intended.

**To Reproduce**
First ensure you have setup `AutoLoginGuards` on all routes. Personally, I did something similar to the following:

```typescript
const appRoutes: Routes = [
  {path: '', pathMatch: 'full', redirectTo: 'home'},
  {path: 'unauthorized', component: UnauthorizedComponent},
  {
    path: 'home',
    canActivate: [AutoLoginGuard],
    loadChildren: () => import('./modules/home/home.module').then(m => m.HomeModule)
  }
  {path: '**', component: PageNotFoundComponent},  // Wildcard route for a 404 page
];
```

Also ensure silent renew is setup:

```typescript
export function configureAuth(oidcConfigService: OidcConfigService) {
  return () =>
    oidcConfigService.withConfig({
      stsServer: environment.identity_url,
      redirectUrl: window.location.origin,
      postLogoutRedirectUri: window.location.origin,
      clientId: environment.client_id,
      scope: environment.scope, // must have offline_access scope
      responseType: 'code',
      silentRenew: true,
      useRefreshToken: true,
      logLevel: LogLevel.Debug,
    });
}
```

1. Open the WebApp
2. Clear all site data 
  ![image](https://user-images.githubusercontent.com/2610199/120673188-928fdd00-c458-11eb-81f5-1fe7d175f5ca.png)
3. Refresh the page (which should cause the user to be authenticated)
4. Open the console and notice silent renew is not running
![image](https://user-images.githubusercontent.com/2610199/120673810-22358b80-c459-11eb-8de2-f04df1abd4e2.png)
5. Wait for the login to timeout and then navigate to another secured route. This should force another sign in.
6. Once the second authentication process has completed, you should be able to open the console and see the silent renew running. If the silent renew process does not start, I've had luck with clearing the cache and refreshing multiple times.
![image](https://user-images.githubusercontent.com/2610199/120675373-9e7c9e80-c45a-11eb-8ac4-5bd76ebb7ce9.png)

**Expected behavior**
[...]

PR #1137, opened 2021-06-14
fix: Start session checks in auto-login guard when already authenticated

liamkeaton:
Suggested fix for #1107 where silent renew was not being performed as
the session checks are only started when `CheckAuthService::checkAuth`
is called. Making `CheckAuthService::startCheckSessionAndValidation`
public is not ideal but reduces duplication of this logic.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login.guard.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login.guard.spec.ts
@@ -78,6 +78,18 @@ describe(`AutoLoginGuard`, () => {
     );
 
     it(
+      'should call startCheckSessionAndValidation() if authenticated already',
+      waitForAsync(() => {
+        spyOn(authStateService, 'areAuthStorageTokensValid').and.returnValue(true);
+        const checkAuthServiceSpy = spyOn(checkAuthService, 'startCheckSessionAndValidation').and.returnValue();
+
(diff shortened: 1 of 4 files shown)


FabianGosebrink:
Thanks for the PR. We would start the silent renew kickoff every time the user hits the route and is authenticated or presses refresh in the browser on a guarded route, is that correct?

liamkeaton:
Yes it tries to start the periodic session check and validation in `PeriodicallyTokenCheckService::startTokenValidationPeriodically` but if already running would return at this point https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/callback/periodically-token-check.service.ts#L35. From some initial tests it does not seem to have any adverse effects as everything in `CheckAuthService::startCheckSessionAndValidation` has `isRunning` checks.

damienbod:
hi @liamkeaton Thanks for the PR!

I'm not sure if this is good, need to understand the auto-login better. @FabianGosebrink @JerrettDavis What's your opinion here?

Greetings Damien

PS we will probably move the V12 branch to main which has a few changes.

liamkeaton:
Thanks @damienbod & @FabianGosebrink for looking at my PR, I've seen that https://github.com/damienbod/angular-auth-oidc-client/pull/1144 should fix the issues I've seen so my suggestion can be closed now. Any idea when this will be released?

FabianGosebrink:
Thanks for this comment. We are currently in a testing phase. And will release afterwards. The next weeks I think :)

PR #1174, opened 2021-07-07
Implemented CanActivateChild interface for auto-login guards

marklagendijk:
With this the guards can be used as argument for `canActivateChild`, so that the guard will be executed when loading any child route:
```ts
export const appRoutes: Routes = [
  {
    path: '', canActivateChild: [AutoLoginAllRoutesGuard], children: [
      {path: '', component: HomeComponent, pathMatch: 'full'},
      {path: 'fetch-data', component: FetchDataComponent},
    ],
  },
  {path: 'unauthorized', component: UnauthorizedComponent, pathMatch: 'full'},
]
```


marklagendijk:
Note: even though it is called `CanActivateChild` it is actually called whenever any descendant is activated, so it also works for deeply nested routes.

FabianGosebrink:
Thanks for this. We will review and merge in the next days! Thanks! Much appreciated.

FabianGosebrink:
Can you take a look at the conflicts here? Thanks :)

marklagendijk:
@FabianGosebrink the conflict is resolved.

FabianGosebrink:
Wonderful, thanks. Will be in the next release.

jacksoncastro:
> Wonderful, thanks. Will be in the next release.

When will the next release be?

FabianGosebrink:
@damienbod is on holiday right now and we have to drive some tests. So I think at the end of next week somewhat around this.

PR #1175, opened 2021-07-07
Popup: calculate top and left position to center the window. Export more services to public API. Add AutoRedirectRoutesGuard

linjie997:
1. Popup: Calculate top and left positions to center the popup window

2. Export to public API: AutoLoginService, AuthStateService, LoginService, ConfigurationProvider.
Exporting such services to public API, can allow the creation of custom RoutesGuard (other than AutoLoginAllRoutesGuard and AutoLoginPartialRoutesGuard).

3. Add AutoRedirectRoutesGuard: Basically it's a copy of AutoLoginPartialRoutesGuard but does not try to login and only saves the requested route.
This may be useful when we want to conditionally trigger the login only later after some checks.


linjie997:
Export to public API: AutoLoginService, AuthStateService, LoginService,
ConfigurationProvider.
Exporting such services to public API, can allow the creation of custom
RoutesGuard (other than AutoLoginAllRoutesGuard and
AutoLoginPartialRoutesGuard).

linjie997:
Sorry for the mess, I will use different branches for each pull request next time.

FabianGosebrink:
We have to be careful to what we expose. We can not expose everything, therefore we have the facade we are providing. Exposing everything would open the door for uncontrolled usecases. Can we split this up? Make on PR for the Popup changes. Then we can merge this. One for the rename then.

linjie997:
Sure, will split up.

PR #1179, opened 2021-07-09
Popup: calculate top and left position to center the window.

linjie997:


FabianGosebrink:
Please fix tests. Thanks!

damienbod:
@linjie997 Thanks for the PR!

Could you fix the unit tests and we can merge?

Greetings Damien

linjie997:
The tests are fixed.
The build and deployment are stuck because of the azure api token, can we skip that?

FabianGosebrink:
Yes that is fine. We have to activate the secret sharing.

PR #1181, opened 2021-07-09
Add AutoRedirectRoutesGuard.

linjie997:
Basically it's a copy of AutoLoginPartialRoutesGuard but does not try to login and only saves the requested route.
This may be useful when we want to conditionally trigger the login only later after some checks.


FabianGosebrink:
I do not get the sense behind this to be honest. What is it good for? You access a route, then the guard triggers and saves the route, when you're not authenticated, then the app - not the lib -  redirects to the STS, you come back , you manually hit the route with the guard again and then the guard redirects you to the saved route. Why not routing to the saved route in the first place?

linjie997:
In that case, is it possible to export only AutoLoginService so I can access to the saved route?

FabianGosebrink:
What you can do now is reading the route from the localstorage. It is stored under `redirect`. But let me understand what your use case is. Maybe we really need to do something. Or maybe we can solve your use case with a guard we already have.

linjie997:
Sure, I'll try to explain my case.

I have this route, which access let's say a Item.
The Item might need or not need authentication depending of its configuration.

If it does not need authentication, it's the simpliest case and I can access the page freely.
If it does need authentication, I have to save the route, start the login flow, then after the successfull login redirect to the previous Item page.

I hope I did explain correctly what I need to do.

FabianGosebrink:
This is exactly what our Route Guards do :) They save the route, start the login flow, redirect you back and then navigate to the stored route. That should cover your case, shouldn't it?

linjie997:
For my second usage case, yes. But in the first case, I might not need to start the login flow for the same route,
If I understand correctly, with the route guard, it will start the authentication flow always right?

FabianGosebrink:
Yes, the Login Guards start the auth flow automatically. But I am interested in the first use case. 

1) You navigate to route `/users`, `/users` has the new `AutoRedirectRoutesGuard`
2) The `AutoRedirectRoutesGuard` saves the `/users` route (?) because you're not authenticated (?)
3) Someone has to kick off the auth, or no? Do you do this? Or don't you want to authenticate?
4) If someone has triggered the auth you get back to your app
5) You navigate to the secured route

This is what the currently existing guards do. Can you help me to understand the first use case? What are you trying to do?

linjie997:
Let me try to explain a real usage case:

I have a list of `Item`.
Each `Item` has its own route `/item/:id`
The item might require autentication to be accessed based on its configuration, let's say each item has a property `isPublic`
When I access to the `/item/:id` route I make an HTTP request to the server;
If the `item` has the `isPublic` property to `true` I receive the `item`,
if `isPublic` is `false` I receive a `401` http response code and I have to trigger the authentication flow.

The "issue" is that after I trigger the authentication flow, the user gets redirected to the initial page and I would like to redirect it to the original `/item/:id` route

FabianGosebrink:
Ah, okay, that is making it clearer. So what you need is an authentication trigger based on a property, which is `isPublic` in your case. If `isPublic` is set to `true` you do not need authentication, otherwise you need authentication. Then this guard in the PR here does not help you I think. You would need to resolve the `isPublic` property in the guard which is a very custom logic. One the one hand because the property is very custom on the other because you need to load data to decide if you need the login or not. The guard in the PR here just saves a route. 

I think a guard is not the correct tool to solve this issue I think. I would solve this with a component which loads the item and does the redirection the item itself (`isPublic = true`) or calls `authenticate` of the lib.

What do you think?

linjie997:
I think the redirection from the component you suggested is a cleaner way instead of the guard.
I guess we can close this PR then, thank you!

Issue #1192, opened 2021-07-14
prompt parameter sent multiple times with silent renew (azure integration)


pfeigl:
**Describe the bug**
When integrating with azuread and using silent renew, when the renew is actually executed, the prompt parameter is sent with multiple values.

MS doesn't allow this and gives an error:
> AADSTS9000411: The request is not properly formatted. The parameter 'prompt' is duplicated.

**To Reproduce**
1. Integrate with AzureAD
2. Enable silent renew
3. wait for the silent renew to happen

**Expected behavior**
prompt is only sent once - however i'm actually not sure what value needs to be sent of the two in this case.

**Screenshots**
![image](https://user-images.githubusercontent.com/237891/125586864-c07fdab9-c1a1-4781-b7b7-62000e0c82be.png)

**Desktop (please complete the following information):**
 - OS: win10
 - Browser chrome
 - Version  91.0.4472.124 

**Additional context**
The reason is, that the schematic for the azure integration has a prompt pre-defined:
https://github.com/damienbod/angular-auth-oidc-client/blob/9cf4ce4e97b574a1c6a78a51cde7232015ea0f7f/projects/schematics/src/ng-add/actions/configs.ts#L38

But also the `createAuthorizeUrl` here always blindly appends the prompt, when one is provided:
https://github.com/damienbod/angular-auth-oidc-client/blob/3de72d57b53f479f946d30b9b77a5bfa975aec79/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts#L332

As for silent-renew the prompt is always defined with the value `none` - see here:
https://github.com/damienbod/angular-auth-oidc-client/blob/3de72d57b53f479f946d30b9b77a5bfa975aec79/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts#L388

This results in URLs like the following
`...&prompt=none&prompt=select_account&...`

Probably the right fix is not to append the prompt but use set (https://angular.io/api/common/http/HttpParams#set)

FabianGosebrink:
Hey, thanks for pointing this out. I see in the `createAuthorizeUrl` method we apply all params no matter if they have been applied before. This is a mistake and a bug imho. Do I understand this correctly? Thanks.

pfeigl:
[...]

PR #1193, opened 2021-07-14
Added fix overwriting param

FabianGosebrink:
#1192

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/check-auth.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/check-auth.service.spec.ts
@@ -436,6 +436,72 @@ describe('CheckAuthService', () => {
         });
       })
     );
+
+    it(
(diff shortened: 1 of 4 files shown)


pfeigl (commented):
Did you intentionally only fix the prompt now and not also all the other appends (scope, nonce, hd)?

FabianGosebrink:
I have to speak to @damienbod or look in the docs if overwriting the params like `nonce`, `scope` etc is even allowed.

pfeigl:
Scope might even be a special case in terms of it beeing kind of list already (space delimited string). Here overwriting probably makes no sense here, however having two scope parameters defined isn't better IMHO.

Perhaps your approach is best by fixing this specific issue, because prompt is the only `customParamsAuthRequest` parameter which is actually used out there right now.

FabianGosebrink:
Have to fix the coverage. @damienbod is in vacation until next week, we will verify this and then merge. Stay tuned here if you want. :)

damienbod:
You could also overwrite the scope maybe?

Issue #1189, opened 2021-07-13
Expose PopupService back in v12


normandjulian:
**Is your feature request related to a problem? Please describe.**
Hi,
In my team, we have severals projects with oidc and angular-auth-oidc-client. It work nicely.

However one of our project is an Office Excel add-in and we have to use Office function to open a dialog : [Office.context.ui.displayDialogAsync](https://docs.microsoft.com/en-us/javascript/api/office/office.ui?view=excel-js-preview#displayDialogAsync_startAddress__options__callback_)

In `angular-auth-oidc-client@11.x.x`  the `PopupService` was exposed.
It allow us to provide our own service instead which use the Office.context.ui.displayDialogAsync instead of window.open

But in `angular-auth-oidc-client@12.x.x` the `PopupService` isn't exposed anymore

**Describe the solution you'd like**
If it's possible to expose it back or to choose if we want to use window.open of something else as **Office.context.ui.displayDialogAsync** in my case

**Describe alternatives you've considered**
Forking the project

**Additional context**
Here is some of my codes 

My office-popup.service.ts
```typescript
@Injectable()
export class OfficePopUpService implements IPopUpService {
    // ...
    private connectDialog: Office.Dialog

    async openPopUp(url: string, popupOptions?: PopupOptions): Promise<void> {
        // ...

        await Office.context.ui.displayDialogAsync(url, optionsToPass, (result) => {
            if (result.status === Office.AsyncResultStatus.Failed) {
                console.log(`${result.error.code} ${result.error.message}`)
            } else {
                this.connectDialog = result.value
                this.connectDialog.addEventHandler(
                    Office.EventType.DialogEventReceived,
                    handleDialogEvent
                )
            }
        })

        // ...
    }
}
```

app.module.ts in v11: 

```typescript

/*
    from the file => angular-auth-oidc-client.d.ts  
    export { PopUpService as ɵbm } from './lib/login/popup/popup.service';
*/
import { ɵbm } from 'angular-auth-oidc-client'

@NgModule({
    declarations: [ /* Some Components */ ],
    imports: [ /* Some imports */ ],
    providers: [
        { provide: ɵbm, useClass: OfficePopUpService },
    ],
    bootstrap: [AppComponent],
})
export class AppModule {}
```

app.module.ts in v12: 

```typescript

/*  not anymore the file => angular-auth-oidc-client.d.ts */
[...]

PR #1199, opened 2021-07-16
Expose PopupService and PopupOptions as public

normandjulian:
Hi,

Related to the issue #1189 , I add the PopupService and PopupOptions in the public export


damienbod:
LGTM thanks for the PR

damienbod:
Just as a note: We just merged the PR and did a release, but the APIs we are providing are supported and we believe overwriting the PopupService might lead to problems using the lib and we do not recommend this.

PR #1207, opened 2021-07-22
Expose PopupService as public #1199

normandjulian:
I'm sorry for the last PR, I exposed the wrong service `PopUpLoginService` instead of `PopUpService` ...


FabianGosebrink:
Hey, thanks, do we need to expose the `PopupLoginService ` then?

normandjulian:
Not really you're right, I'll update my PR

FabianGosebrink:
Can you please update this PR accordingly, that we can merge? Thanks.

normandjulian:
It should be good now ! Thank you.

Issue #1170, opened 2021-07-05
id_token nonce validation fails randomly in 12.0.0


maxgr0:
**Describe the bug**
As there were some issues with the validation of id tokens already here, I think the behaviour is known. I'am using this library at v12.0.0 and Angular 12.1.1. Sometimes, the nonce validation fails as the values don't match. I think probably some kind of regression happened here when going from 11.x to 12.0.0. 
I've logged all tokens on backend-side to check if it's the error of our oidc implementation but that's not the case as tokens with the correct nonce are issued.
Following scenario happens:
- Library asks for a token, get's one back
- Immediately does subsequent requests but validates the first token and fails with an error stating that the nonce is invalid.
- The behaviour is exactly the same as on old issues but after 3-10 requests it works and the user is signed in.
- I've already increased the `maxIdTokenIatOffsetAllowedInSeconds` to 600.

**To Reproduce**
Use https://github.com/damienbod/angular-auth-oidc-client/tree/main/projects/sample-code-flow-auto-login-all-routes with Angular 12 and v12.0.0 of this library.

**Expected behavior**

**Screenshots**
![Screenshot 2021-07-06 at 12 16 11 AM](https://user-images.githubusercontent.com/48564150/124521518-09b9e780-ddf0-11eb-87b2-5558667a5462.png)

**Desktop (please complete the following information):**
 - OS: Mac OS 11.2.3
 - Browser: Tested with Safari & Chrome

maxgr0:
@FabianGosebrink @damienbod any updates on that? I would be more than happy to provide a PR but currently don't really know where to start.
Cheers!🍻

dernilz:
A while ago i had the same problem using Keycloak as OIDC server.

Via [documentation](https://nice-hill-002425310.azurestaticapps.net/docs/documentation/configuration#ignorenonceafterrefresh) I stumbled apon this config option:

> ### `ignoreNonceAfterRefresh`
> 
> - Type: boolean
> - Required: false
> 
> A token obtained by using a refresh token normally doesn't contain a nonce value.   
> The library checks it is not there. However some oidc endpoint implementations do send one.   
> Setting ignoreNonceAfterRefresh to true disables the check if a nonce is present.   
> Please note that the nonce value, if present, will not be verified. Default is false.   

[...]

PR #1214, opened 2021-07-26
Additional logging when a nonce is created and validated

jochenjonc:
Added some additional logging to help in figuring out race conditions when authenticating, see #1170


damienbod:
Thanks for the PR

I think the nonce it logged to many times now. Maybe inside the function is enough?

Greetings Damien

jochenjonc:
Hi @damienbod,

I've logged it everywhere I thought it could be useful, but indeed in the createNonce function should be enough.
I'll push the changes.

Regards,
Jochen

FabianGosebrink (approved):

damienbod:
Thanks @jochenjonc 

Greetings Damien

PR #1217, opened 2021-07-29
Move and export PASSED_CONFIG, PassedInitialConfig

linjie997:
Move PassedInitialConfig interface to its own file.
Create passed-config folder.

Move PASSED_CONFIG InjectionToken to passed-config folder.

Export PassedInitialConfig interface from Public API file since it's a public interface.
We can can now import it like this:
`import { PassedInitialConfig } from 'angular-auth-oidc-client'`
instead of
`import { PassedInitialConfig } from 'angular-auth-oidc-client/lib/auth.module'`

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/angular-auth-oidc-client.ts
+++ b/projects/angular-auth-oidc-client/src/lib/angular-auth-oidc-client.ts
@@ -10,6 +10,8 @@ export * from './config/auth-well-known/auth-well-known-endpoints';
 export * from './config/config.service';
 export * from './config/loader/config-loader';
 export * from './config/openid-configuration';
+export * from './passed-config/passed-config';
+export * from './passed-config/passed-initial-config';
 export * from './interceptor/auth.interceptor';
 export * from './logging/log-level';
 export * from './logging/logger.service';
--- a/projects/angular-auth-oidc-client/src/lib/auth.module.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/auth.module.spec.ts
@@ -1,6 +1,7 @@
 import { TestBed, waitForAsync } from '@angular/core/testing';
 import { of } from 'rxjs';
-import { AuthModule, PASSED_CONFIG } from './auth.module';
+import { AuthModule } from './auth.module';
+import { PASSED_CONFIG } from './passed-config/passed-config';
 import { OidcConfigService } from './config/config.service';
 import { OidcConfigServiceMock } from './config/config.service.mock';
 import { StsConfigHttpLoader, StsConfigLoader, StsConfigStaticLoader } from './config/loader/config-loader';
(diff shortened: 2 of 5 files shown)


FabianGosebrink:
Hey, thanks for this Pull Request. Why do we have to expose the PassedInitialConfig?

linjie997:
Hello, since we could use a provider for the PASSED_CONFIG, it's common to use a factory function to provide it.
Let me write down an example:

```
imports: [
     AuthModule.forRoot({})
],
providers: [
    {
      provide: PASSED_CONFIG,
      useFactory: getPassedConfig,
    },
]

.....

// This is a simple mock config, we might use an async method to get the configuration
function getPassedConfig(): PassedInitialConfig {
  return {
    config: {
      authority: 'https://authority.com/sample',
      redirectUrl: `${location.origin}${location.pathname}`,
      clientId: 'clientId',
      silentRenewUrl: '/silent-renew.html',
      scope: 'openid profile email',
      responseType: 'code',
      postLogoutRedirectUri: `${location.origin}`,
      ignoreNonceAfterRefresh: true,
      silentRenew: true,
      useRefreshToken: true,
    },
  };
}
```


Without directly exposing the PassedInitialConfig interface, the import is now looking like this:

```
import { AuthModule, PASSED_CONFIG } from 'angular-auth-oidc-client';
import { PassedInitialConfig } from 'angular-auth-oidc-client/lib/auth.module';
```

I think it would be better and cleaner to expose it so the import will look like this:

```
import { AuthModule, PASSED_CONFIG, PassedInitialConfig } from 'angular-auth-oidc-client';
```

What do you think about it?

FabianGosebrink:
I think the configuration is being misused here. You can either use the `StsConfigHttpLoader` or the ` StsConfigStaticLoader` to pass your config. Passing the `PASSED_CONFIG` should not be a use case. What is your situation? Are you getting the config from a service call in the APP_INITIALIZER?

linjie997:
Yes I do get the configuration with an async call on APP_INITIALIZER.
I checked the doc and I think I could use StsConfigHttpLoader, however do you think it would be a viable alternative to pass the configuration with the provider as the example I suggested?

FabianGosebrink:
We designed the loaders to exactly cover this issue. But we did not document it, as far as I can see. Will add a documentation for it. 

```
imports: [AuthModule.forRoot({ loader: { provide: StsConfigLoader, useFactory: configureAuth, deps: [ConfigService], }, });]
```

Something like this. Does this help?

linjie997:
This should be working for my case, I'll try ASAP and provide feedback here.
Thanks!

FabianGosebrink:
Thank you!

linjie997:
Hey Fabian, I can confirm I got it working starting from your configuration sample.

My configuration is something like this:

```
import { AuthModule, StsConfigLoader, StsConfigStaticLoader } from 'angular-auth-oidc-client';

imports: [
    AuthModule.forRoot({
      loader: {
        provide: StsConfigLoader,
        useFactory: getConfig,
        deps: [MyService],
      },
    }),
]

................................................................................

export function getConfig(service: MyService) {
  const config: OpenIdConfiguration = service.getConfig();
  return new StsConfigStaticLoader(config);
}
```

I think we can close this PR since it's not needed, thank you for your quick responses!

FabianGosebrink:
Thank you for posting the sample! I will add it to the docs as this question arose a few times now. Thank you for your collaboration :)

PR #1231, opened 2021-08-17
docs(guards): use UrlTree for redirect, clean up

fmalcher:
- use `UrlTree` as the guard return type for the redirect
- clean up the example: Since the docs should teach the concept, I tried to make the method as  understandable as possible.
- revised and extended text


FabianGosebrink:
Thanks!

PR #1237, opened 2021-08-19
Fix browser storage issues

FabianGosebrink:


projects/angular-auth-oidc-client/src/lib/storage/browser-storage.service.ts
```diff
@@ -14,32 +14,40 @@ export class BrowserStorageService {
...
-      this.loggerService.logDebug(configId, `Wanted to write '${key}/${value}' but Storage was falsy`);
+      this.loggerService.logDebug(configId, `Wanted to write '${value}' but Storage was falsy`);
 
       return false;
     }
 
     value = value || null;
 
-    storage.write(`${key}`, JSON.stringify(value));
+    storage.write(`${configId}`, JSON.stringify(value));
```
fmalcher, line 50: Is there a reason why this is not just like this?

```suggestion
    storage.write(configId, JSON.stringify(value));
```

FabianGosebrink: Hey, thanks for the review. The PR is still in WIP and I wanted to look at some things anyway. But this is a nice catch which I will change for sure! Thanks!

damienbod:
LGTM

PR #1270, opened 2021-10-06
docs: update the example function name

dalejodc:
The example is showing the _logout_ process but the function mane is "_login_"


damienbod:
LGTM, thanks

PR #1307, opened 2021-11-21
Fabiangosebrink/remove app initializer

FabianGosebrink:
We do not want to use the APP_INITIALIZER anymore to not confuse or affect the bootstrapping process of the users' app. The config should be set through the loaders and things should kick off when the user interacts with the lib for the first time, like calling a method.

If the user wants to speed up the process, we want to provide a method where you can preload the auth well known document explicitly, so then it is set and not has to be loaded on the start.

- [x] docs
- [x] testing
- [x] migration guide


docs/site/angular-auth-oidc-client/docs/documentation/public-api.md
```diff
@@ -164,6 +164,19 @@ The `stsCallback$` observable emits _after_ the library has handled the possible
 this.stsCallback$ = this.oidcSecurityService.stsCallback$;
 ```
 
+## preloadAuthWellKnownDocument(configId?: string)
+
+With this method you can explicitly load your auth well-known document to persist it in the configured storage of the lib. This method will perform an http call to load and then save the information. If you do not call this method explicitly, the lib will call it when it is needed for the first time automatically.
+
```
damienbod, line 170: This method allows you to explicitly load the secure token servers well-known endpoints and persist the settings to the configured storage of this library. This method will perform an http call to load and save the information. If you do not call this method explicitly, the lib will call it when it is needed for the first time automatically.

docs/site/angular-auth-oidc-client/docs/migrations/v13-to-v14.md
```diff
@@ -0,0 +1,185 @@
...
+In this version we removed the `APP_INITIALIZER` ([See PR](https://github.com/damienbod/angular-auth-oidc-client/pull/1307)).
+The library now does nothing until you interact with her. No loading of anything and it does not affect your application's bootstrapping process at all.
+You can however explicitly preload your Auth Well-Known Endpoints with a new method called [preloadAuthWellKnownDocument()](/docs/documentation/public-api#preloadauthwellknowndocumentconfigid-string). As a side effect becase the config has to be loaded first, a lot of APIs become reactive and return an Observable now.
+
+All changes are describes below.
+
+## `eagerLoadAuthWellKnownEndpoints` was removed
+
+The Auth Well Known Endpoints are always eager loaded expect loading it explicitly using the [preloadAuthWellKnownDocument()](/docs/documentation/public-api#preloadauthwellknowndocumentconfigid-string) method. No option needed anymore.
+
```
damienbod, line 16: # Version 13 to 14

In this version the `APP_INITIALIZER` was removed. ([See PR](https://github.com/damienbod/angular-auth-oidc-client/pull/1307)).
The library will not do anything until the application interact with it. There is no pre-loading of anything and it does not affect your application's bootstrapping process at all.
You can however explicitly preload the secure token server well-known endpoints with a new method called [preloadAuthWellKnownDocument()](/docs/documentation/public-api#preloadauthwellknowndocumentconfigid-string). As a side effect because the config has to be loaded first, a lot of APIs become reactive and return an Observable now.

All changes are described below.

## `eagerLoadAuthWellKnownEndpoints` was removed

The secure token server well known endpoints are always eager loaded expect loading it explicitly using the [preloadAuthWellKnownDocument()](/docs/documentation/public-api#preloadauthwellknowndocumentconfigid-string) method. No option needed anymore.

PR #1342, opened 2022-01-03
refresh token rotation should be optional (fixes #1338)

### LINKED ISSUE #1338: [Bug]: Error refreshing access token (OP without refresh token rotation)
*Description:*

After 2nd access token refresh, refreshing stops with the error :
`[ERROR] - no refresh token found, please login`

---

version: 13.1.0
version Angular: 13.1.0
browser: Firefox, Chrome, Chrome Canary (Windows 10)
OIDC server: MS Active Directory Federation Services (ADFS) 2019
flow: authorization code

---

I am evaluating angular-auth-oidc-client for our angular applications authenticating with ADFS 2019 and having the following configuration:

```
{
  ...
  scope: 'openid profile offline_access',
  responseType: 'code',
  silentRenew: true,
  silentRenewUrl: `${window.location.origin}/assets/silent-renew.html`,
  renewTimeBeforeTokenExpiresInSeconds: 30,
  useRefreshToken: true,  
  ...
}
```

The refresh token lifetime is longer than the expiration of the access token (access token expiration: 1h / refresh expiration: 24h).

The first refresh is OK, but during the second refresh I get the following error:

error-js-01

In my opinion the problem is that when refreshing the access token, angular-auth-oidc-client looses the refresh token, ADFS 2019 does not rotate refresh tokens (if valid) and does not send with the refresh response a new refresh token.
The next time the token is refreshed there's no refresh token and I get the error.

It looks like the OP behavior (ADFS) is normal, in fact the rotation of the refresh token is optional as specified by The OAuth 2.0 RFC:

[rfc6749, 1.5 Refresh Token, step H](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5)
"The authorization server authenticates the client and validates the refresh token, and if valid, issues a new access token (and, **optionally**, a new refresh token)."

---

Here's a more precise description of Client/OP interactions:

1. **Getting token**

  - request:

    GET .../token
    content: grant_type=authorization_code&client_id=07...a3913&code_verifier=e4...Am12Z&code=iS...vBJAA&redirect_uri=http%3A%2F%2Flocalhost%3A4200

  - response:

```
    {
      "access_token":"ey...UKDPg",
      "token_type":"bearer",
      "expires_in":3600,*
      "resource":"07...a3913",
      "refresh_token":"iV...mdL6g",
      "refresh_token_expires_in":28800,
      "scope":"profile openid",
      "id_token":"eyJ...D8nMw"
    }
```    

2. **1st refresh:**

  - request:

      GET .../token
      content: grant_type=refresh_token&client_id=07...a3913&refresh_token=iV...mdL6g

  - response:

     {"access_token":"ey...ONA","token_type":"bearer","expires_in":3600,"id_token":"ey...hw"}

**2nd refresh:**

Error JS

mduriancik:
Proposed solution to #1338. The idea is to separate the refresh token from authResult in the storage. 
I have tested the solution and it resolves my problem.

Of course more refactoring and more tests could be needed.

### DIFF
--- a/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
+++ b/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
@@ -465,3 +465,13 @@ Default = _3_. <br/>
 - Required: `false`
 
 Adds the `ngsw-bypass` param to all requests ([Angular Documentation](https://angular.io/guide/service-worker-devops#bypassing-the-service-worker)).
+
+### `allowUnsafeReuseRefreshToken`
+
+- Type: `boolean`
+- Required: `false`
+
+Allows the refresh token reuse. Refresh tokens are typically longer-lived and RFC6749 allows their reuse nevertheless this is strongly discourage.
+Activate this property only if your OIDC provider cannot be configured to rotate refresh tokens.
+
+Default = _false_
--- a/projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
+++ b/projects/angular-auth-oidc-client/src/lib/config/openid-configuration.ts
@@ -172,4 +172,8 @@ export interface OpenIdConfiguration {
   refreshTokenRetryInSeconds?: number;
   /** Adds the ngsw-bypass param to all requests */
   ngswBypass?: boolean;
+  /** Allow refresh token reuse (refresh without rotation), default value is false.
+   * The refresh token rotation is optional (rfc6749) but is more safe and hence encouraged.
+   */
+  allowUnsafeReuseRefreshToken?: boolean;
 }
--- a/projects/angular-auth-oidc-client/src/lib/flows/callback-context.ts
+++ b/projects/angular-auth-oidc-client/src/lib/flows/callback-context.ts
@@ -18,6 +18,8 @@ export interface AuthResult {
   id_token?: string;
   // eslint-disable-next-line @typescript-eslint/naming-convention
   access_token?: string;
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  refresh_token?: string;
   error?: any;
   // eslint-disable-next-line @typescript-eslint/naming-convention
   session_state?: any;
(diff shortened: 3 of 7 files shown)


damienbod:
Hi @mduriancik  firstly thanks for the PR. I'm thinking about this. I think we need an **allowUnsafereUseRefreshToken: bool** configuration to activate this feature. Secondly not sure about having the refresh token saved twice to the session storage. 

Not sure of a solution at present. 

Greetings Damien

FabianGosebrink:
Any news on this?

mduriancik:
Hi Fabien,

Thank's for your feedback. Yes, good idea, I can add the property allowUnsafeReuseRefreshToken (default true?).

Yes it is not quite "DRY" to store the refresh token twice (authResult and refresh_token), but as refresh_token can outlive the immediate authResult it looks to me the simplest solution.

Other solution: to modify the new authResult transfering the refresh_token form the previous authResult -> It looks to me less "elegant", but if you prefer this one I can change my PR.  Do you see there other possibilities?

Milan

damienbod:
allowUnsafeReuseRefreshToken (default false?). We do not want to encourage this, this should only be used if the IDP is missing the feature. This is not a good idea and I'm not even sure we should support this. Maybe ADFS federation to AAD for public IDP  would be better. 

Greetings Damien

Greetings Damien

mduriancik:
added allowUnsafeReuseRefreshToken
a little bit of refactoring:
- the storage property renamed: reusable_refresh_token -> to clarify the its purpose of this token
- default refresh token location is in 'authnResult'
- refresh token is stored in "reusable_refresh_token" only if the property is enabled

damienbod:
Hi @mduriancik I will add this to the next release , Thanks!

Greetings Damien

sonzo:
What is the status on this?

i can see the build didnt go through
https://github.com/damienbod/angular-auth-oidc-client/runs/4779278390?check_suite_focus=true

PR #1353, opened 2022-01-14
⌛isLoading observable in OidcSecurityService

projectaki:
Hi, I made the pull request to add an isLoading observable to the oidc security service. 

**Reason** 

I came accross the problem when all routes are protected by an Auth guard. The Auth guard subscribes to the isAuthenticated observable in the oidc security service. If emission is true return true, otherwise call the authorize method. Since the initial emission of isAuthenticated is false, this would cause an infinite redirect loop. In these situations it could be convenient to have an isLoading observable, which is implemented to emit false when the checkAuth methods have either emitted, or errored.

**Use case**

Combine isLoading with isAuthenticated to avoid timing issues, isAuthenticated will only emit its first value after isLoading has emitted false.

Ps: It is the best auth library I have used.<3

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
@@ -633,127 +633,142 @@ describe('OidcSecurityService', () => {
   });
 
   describe('test isLoading$', () => {
-    it('should emit true', () => {
-      oidcSecurityService.isLoading$.subscribe((x) => {
-        expect(x).toBeTrue();
-      });
-    });
+    it(
+      'should emit true',
+      waitForAsync(() => {
+        oidcSecurityService.isLoading$.subscribe((x) => {
+          expect(x).toBeTrue();
+        });
+      })
+    );
 
-    it('isloading should emit false after checkauth is called', (done) => {
-      const config = { configId: 'configId1' };
-      spyOn(configurationService, 'getOpenIDConfigurations').and.returnValue(of({ allConfigs: [config], currentConfig: config }));
-      spyOn(checkAuthService, 'checkAuth').and.returnValue(of(null));
+    it(
+      'isloading should emit false after checkauth is called',
+      waitForAsync(() => {
+        const config = { configId: 'configId1' };
+        spyOn(configurationService, 'getOpenIDConfigurations').and.returnValue(of({ allConfigs: [config], currentConfig: config }));
+        spyOn(checkAuthService, 'checkAuth').and.returnValue(of(null));
 
-      oidcSecurityService
-        .checkAuth()
-        .pipe(
-          switchMap(() =>
-            oidcSecurityService.isLoading$.pipe(
-              tap((x) => {
-                expect(x).toBeFalse();
-                done();
-              })
+        oidcSecurityService
+          .checkAuth()
+          .pipe(
+            switchMap(() =>
+              oidcSecurityService.isLoading$.pipe(
+                tap((x) => {
+                  expect(x).toBeFalse();
+                })
+              )
             )
           )
-        )
-        .subscribe();
-    });
+          .subscribe();
+      })
+    );
 
-    it('isloading should emit false on error in checkauth', (done) => {
-      const config = { configId: 'configId1' };
-      spyOn(configurationService, 'getOpenIDConfigurations').and.returnValue(of({ allConfigs: [config], currentConfig: config }));
-      spyOn(checkAuthService, 'checkAuth').and.returnValue(throwError(() => new Error('Error')));
-      oidcSecurityService
-        .checkAuth()
-        .pipe(
-          catchError(() =>
-            oidcSecurityService.isLoading$.pipe(
-              tap((x) => {
-                expect(x).toBeFalse();
-                done();
-              })
+    it(
+      'isloading should emit false on error in checkauth',
+      waitForAsync(() => {
+        const config = { configId: 'configId1' };
+        spyOn(configurationService, 'getOpenIDConfigurations').and.returnValue(of({ allConfigs: [config], currentConfig: config }));
+        spyOn(checkAuthService, 'checkAuth').and.returnValue(throwError(() => new Error('Error')));
+        oidcSecurityService
+          .checkAuth()
+          .pipe(
+            catchError(() =>
+              oidcSecurityService.isLoading$.pipe(
+                tap((x) => {
+                  expect(x).toBeFalse();
+                })
+              )
             )
           )
-        )
-        .subscribe();
-    });
+          .subscribe();
+      })
+    );
 
-    it('isloading should emit false after checkauthMultiple is called', (done) => {
-      const config = { configId: 'configId1' };
-      spyOn(configurationService, 'getOpenIDConfigurations').and.returnValue(of({ allConfigs: [config], currentConfig: config }));
-      spyOn(checkAuthService, 'checkAuthMultiple').and.returnValue(of(null));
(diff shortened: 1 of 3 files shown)


projects/angular-auth-oidc-client/src/lib/oidc.security.service.ts
```diff
@@ -61,6 +61,15 @@ export class OidcSecurityService {
     return this.callbackService.stsCallback$;
   }
 
+  /**
+   * Emits false when checkAuth observable emits a value, or errors. Initial value: true.
+   */
+  get isLoading$(): Observable<boolean> {
+    return this.isLoadingBehaviorSubject.asObservable();
```
FabianGosebrink, line 68: Please change the name in `isLoading` instead of `isLoadingBehaviorSubject`. The type should not be visible in the name of the variable.

projects/angular-auth-oidc-client/src/lib/oidc.security.service.ts
```diff
@@ -414,4 +429,13 @@ export class OidcSecurityService {
...
       .pipe(switchMap((config) => this.urlService.getAuthorizeUrl(config, customParams)));
   }
+
+  private finishLoading = (): void => {
+    this.isLoadingBehaviorSubject.next(false);
+  };
+
+  private finishLoadingOnError = (err: any): Observable<never> => {
+    this.isLoadingBehaviorSubject.next(false);
+    return throwError(() => err);
```
FabianGosebrink, line 439: newline before return

FabianGosebrink (changes requested):

FabianGosebrink:
Could you add the docs as well? Nice job, thank you!

projects/angular-auth-oidc-client/src/lib/oidc.security.service.ts
projectaki: I have fixed this now.

projects/angular-auth-oidc-client/src/lib/oidc.security.service.ts
projectaki: I have fixed this now.

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,129 @@ describe('OidcSecurityService', () => {
       });
     });
   });
+
+  describe('test isLoading$', () => {
+    it('should emit true', () => {
```
FabianGosebrink, line 636: waitForAsync

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,129 @@ describe('OidcSecurityService', () => {
...
+      oidcSecurityService.isLoading$.subscribe((x) => {
+        expect(x).toBeTrue();
+      });
+    });
+
+    it('isloading should emit false after checkauth is called', (done) => {
```
FabianGosebrink, line 642: replace done with waitForAsync

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,129 @@ describe('OidcSecurityService', () => {
...
+                expect(x).toBeFalse();
+                done();
+              })
+            )
+          )
+        )
+        .subscribe();
+    });
+
+    it('isloading should emit false on error in checkauth', (done) => {
```
FabianGosebrink, line 662: replace done with waitForAsync

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,129 @@ describe('OidcSecurityService', () => {
...
+                expect(x).toBeFalse();
+                done();
+              })
+            )
+          )
+        )
+        .subscribe();
+    });
+
+    it('isloading should emit false after checkauthMultiple is called', (done) => {
```
FabianGosebrink, line 681: replace done with waitForAsync

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,129 @@ describe('OidcSecurityService', () => {
...
+                expect(x).toBeFalse();
+                done();
+              })
+            )
+          )
+        )
+        .subscribe();
+    });
+
+    it('isloading should emit false on error in checkauthMultiple', (done) => {
```
FabianGosebrink, line 701: replace done with waitForAsync

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,129 @@ describe('OidcSecurityService', () => {
...
+                expect(x).toBeFalse();
+                done();
+              })
+            )
+          )
+        )
+        .subscribe();
+    });
+
+    it('isloading should emit false after checkAuthIncludingServer is called', (done) => {
```
FabianGosebrink, line 720: replace done with waitForAsync

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,129 @@ describe('OidcSecurityService', () => {
...
+                expect(x).toBeFalse();
+                done();
+              })
+            )
+          )
+        )
+        .subscribe();
+    });
+
+    it('isloading should emit false on error in checkAuthIncludingServer', (done) => {
```
FabianGosebrink, line 740: replace done with waitForAsync

FabianGosebrink (changes requested):

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,144 @@ describe('OidcSecurityService', () => {
...
+      })
+    );
+
+    it(
+      'should emit false on error in checkauth',
+      waitForAsync(() => {
+        const config = { configId: 'configId1' };
+        spyOn(configurationService, 'getOpenIDConfigurations').and.returnValue(of({ allConfigs: [config], currentConfig: config }));
+        spyOn(checkAuthService, 'checkAuth').and.returnValue(throwError(() => new Error('Error')));
+        oidcSecurityService
```
FabianGosebrink, line 673: new line between spy and `oidcSecurityService` (for the complete test)

projects/angular-auth-oidc-client/src/lib/oidc.security.service.spec.ts
```diff
@@ -631,4 +631,144 @@ describe('OidcSecurityService', () => {
...
+          .checkAuth()
+          .pipe(
+            catchError(() =>
+              oidcSecurityService.isLoading$.pipe(
+                tap((x) => {
```
FabianGosebrink, line 678: We do not need the tap because catch error returns the isLoading, and we can check in the subscribe for the result (complete test)

FabianGosebrink (changes requested):

FabianGosebrink:
Last fixes and then we are good. Thanks for the awesome work.

projectaki:
> Last fixes and then we are good. Thanks for the awesome work.

My pleasure! Thank you for taking the time to review it!

FabianGosebrink (approved):

PR #1359, opened 2022-01-17
Fix `getUrlParameter`'s handling of fragment response

### LINKED ISSUE #1346: [Bug]: First fragment parameter is ignored
**What Version of the libraray are you using?**
I am using 12.0.3, but the error in the implementation of `getUrlParameter` is the same in the 13 branch and the main branch.

**Describe the bug**
`getUrlParameter` in the UrlService does not parse properly oauth request with a response_type of fragment.
( https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts#L26 )
specifically, it will skip the first parameter, so if the oauth provider redirect to `https://mysite.com/#code=abc&state=123&...` , the library will return a `no code in url` error.
This produce a looping behavior that will eventually resolve only when the first fragment is not required in the case of an openid provider returning fragments in an indetermined order, or will never work if the first fragment is always "code" or "state".

**To Reproduce**
Receive an oauth response in `response_mode = 'fragment'` with the first fragment being a required parameter for the oauth flow.

**Expected behavior**
The implementation should properly read the first fragment parameter.
https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts#L26

Samuel-B-D:
This fixes #1346

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.spec.ts
@@ -120,13 +120,16 @@ describe('UrlService Tests', () => {
     });
 
     it('gets correct params when response_mode=fragment', () => {
-      // Note: This is the example access token response from the RFC: https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2
-      const urlToCheck = 'https://www.example.com/signin#state=thisisastate&ui_locales=fr&code=thisisacode';
-      const code = service.getUrlParameter(urlToCheck, 'code');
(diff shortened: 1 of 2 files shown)


Samuel-B-D:
I think a test for square bracket correct handling should also be there for `getUrlParameter` but I am unsure if you would accept it within this PR or if it should be done in another one.

A test for handling of `[]` url parameters seems justified since it's an edge-case which required particular handling in the current implementation, and must be upheld if the implementation were to change.

Issue #1340, opened 2022-01-02
[Refactoring]: Expose required methods through public API


FabianGosebrink:
Not sure if we want to do this, because I do not see the need to validate a token explicitly. The lib should take care of this internally.

`LoggerService`, `OidcConfigService` are candidates which are used but maybe can be exposed through the facade.

FabianGosebrink:
I see the purpose of `OidcConfigService`. We will add it back in V14. Why is the `LoggerService` needed?

timdeschryver:
I was able to remove the `OidcConfigService` import, this was a left-over from an older version.
We have the same implementation of the `LoggerService`, but we're logging the messages to our logging service (not to the console). We also have a filter in place to remove some of the logs e.g. `Error: login_required`.

gbenthomas:
We are using a custom implementation of LoggerService similar to what @timdeschryver mentioned. 

Looks like LoggerService was exposed due to https://github.com/damienbod/angular-auth-oidc-client/issues/487

FabianGosebrink:
Is it okay for you both to just provide your custom implementation, or do you think like an abstract logger which you can overwrite like the storage service would be better?

FabianGosebrink:
Thanks for the feedback. I think we have to implement that and make it to land in V14. FYI. @damienbod

PR #1414, opened 2022-03-25
fix: include query params in return url with canLoad guards

### LINKED ISSUE #1409: [Bug]: Query params are stripped by canLoad guards when user isn't authenticated

v11, but v12 and v13 have the same behavior.

**Describe the bug**

While using the `canLoad` guards the query parameter info is lost when there's no auth state in the storage.
This is because the persisted url only uses the route's segments.
When the authentication state is already persisted, this doesn't happen because the URL doesn't need to be stored.

For example, when a user navigates to `/search?query=hello`, the url `/search` is persisted to storage without the query parameter info `?query=hello`.

**To Reproduce**

1. Guard a route with the `AutoLoginAllRoutesGuard` or `AutoLoginPartialRoutesGuard` guard
2. Open a new browser (or tab), then navigate to that route and append a query parameter
3. The browser redirects and comes back to the page without the query parameter

**Expected behavior**

I expect that the query parameter doesn't get lost in the process.

**Screenshots**

/

- OS: [e.g. iOS] all
- Browser [e.g. chrome, safari] all
- Version [e.g. 22] /

**Smartphone (please complete the following information):**

/

**Additional context**

This bug was partially resolved with #1051, but the wrong redirect URL is stored.
I think a solution would be to use [`router.getCurrentNavigation().toString()`](https://angular.io/api/router/Router#getCurrentNavigation) to get the complete URL (with query parameters), instead of using the segments from the `canLoad` method. This could also be used for the `canActivate` guards to keep both methods consistent.

First, I thought that using `location.path()` could also work, but this returns the current URL, not the URL that's being navigated to.

I found the `getCurrentNavigation` by looking into some Angular issues with a similar problem https://github.com/angular/angular/issues/17359 and https://github.com/angular/angular/issues/30633.

timdeschryver:
I tried the fix in the project that I'm working on, and also in the `start-sample-code-flow-auto-login` example in this repo.

yes

Resolves #1409

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login-all-routes.guard.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login-all-routes.guard.spec.ts
@@ -1,5 +1,5 @@
 import { TestBed, waitForAsync } from '@angular/core/testing';
-import { RouterStateSnapshot, UrlSegment } from '@angular/router';
+import { Router, RouterStateSnapshot } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
 import { Observable, of } from 'rxjs';
 import { mockClass } from '../../test/auto-mock';
@@ -18,6 +18,7 @@ describe(`AutoLoginAllRoutesGuard`, () => {
   let storagePersistenceService: StoragePersistenceService;
   let configurationService: ConfigurationService;
   let autoLoginService: AutoLoginService;
+  let router: Router;
 
   beforeEach(() => {
     TestBed.configureTestingModule({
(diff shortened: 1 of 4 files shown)


timdeschryver:
Before I forget, when I upgraded our project to use the v14 version, I bumped against two problems that aren't documented in the migration guide.

The removed of `LoggerService` and `OidcConfigService`.

FabianGosebrink (approved):

FabianGosebrink:
V14 is not released yet. We are currently looking to which services to expose and which not. Can we discuss this here: https://github.com/damienbod/angular-auth-oidc-client/issues/1340

PR #1420, opened 2022-04-07
fix(ng-add): remove double quotes in info messages

timdeschryver:
I couldn't get enough, so I installed the package in another project.
In the process I saw that the info messages from the ng-add command included two double quotes.

### DIFF
--- a/projects/schematics/src/ng-add/files/auth-http-config/auth-http-config.module.__ts__
+++ b/projects/schematics/src/ng-add/files/auth-http-config/auth-http-config.module.__ts__
@@ -27,8 +27,7 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
           // autoUserInfo: false,
         };
       })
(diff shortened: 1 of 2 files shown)


projects/schematics/src/ng-add/files/auth-http-config/auth-http-config.module.__ts__
```diff
@@ -27,8 +27,7 @@ export const httpLoaderFactory = (httpClient: HttpClient) => {
           // autoUserInfo: false,
         };
       })
-    )
-    .toPromise();
```
timdeschryver, line 31: Removed because this is deprecated, and `StsConfigHttpLoader` now only accepts an Observable.

FabianGosebrink (approved):

FabianGosebrink:
LGTM! Thanks @timdeschryver !

PR #1430, opened 2022-04-22
Reset the silent renew running state when user forces refresh

liamkeaton:
Currently you can only call userForceRefreshSession once as the
silent renew running state is set and not reset. On the second call
the "silent renew process is probably stuck" state is hit which
clears all state and performs a refresh even if the tokens were not
expired.


damienbod:
@FabianGosebrink this might have side effects, need to validate this good.

FabianGosebrink:
I think this makes sense. Can we have a unit test with it, checking if after the call the `this.flowsDataService.resetSilentRenewRunning(config);` has been called?

FabianGosebrink (changes requested):
Okay for me, please add at least one unit test.

FabianGosebrink:
If we have one unit test we can merge imho...

FabianGosebrink:
@liamkeaton are you still up for this PR? Please merge main, add a unit test, and we can merge.

FabianGosebrink:
Nice, thanks for the heads up :)

Issue #1454, opened 2022-06-07
[Feature Request]: Add a way to disable PKCE support


Kwoin:
**Is your feature request related to a problem? Please describe.**
I found no way to disable the code verifier generation in code flow.
An authority with no PKCE support would fail to exchange an authorization code to an authorization token if that authorization code has been generated with a code verifier.

**Describe the solution you'd like**
Add a config parameter to disable PKCE support. Meaning the authorize request will not countain the `code_challenge` and `code_challenge_method` query params.

**Describe alternatives you've considered**
none

**Additional context**
none

PR #1455, opened 2022-06-07
add disablePkce config parameter

Kwoin:
Address issue #1454


projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -488,6 +485,13 @@ export class UrlService {
     );
   }
 
+  private getCodeChallenge(config: OpenIdConfiguration): Observable<string> {
+    if (config.disablePkce) return of(null);
```
FabianGosebrink, line 489: Please spend the return an extra line and curly braces

```
if (config.disablePkce) {
   return of(null);
}
```

projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -488,6 +485,13 @@ export class UrlService {
...
+    // code_challenge with "S256"
+    const codeVerifier = this.flowsDataService.createCodeVerifier(config);
```
FabianGosebrink, line 491: newline before return

FabianGosebrink (changes requested):

projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
Kwoin: done

projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
Kwoin: done

damienbod:
Thanks for the PR. 

I am not sure we should support this as using code flow without PKCE is not a good idea. A better solution would be to add support on the server or change the IDP. 

Greetings Damien

Kwoin:
Thanks for your review,

To better describe my case:
- IDP enforced by stakeholders
- client configured as "confidential" client in IDP
- IDP has no PKCE support for confidential client

I am aware that PKCE should be enabled for confidential as well as public client, but not all idp currently supports it for confidential client.

My current work around is to manually navigates to a hand-crafted URL :

```
          this.oidcSecurityService.checkAuth().subscribe(({isAuthenticated, userData, accessToken, idToken}) => {
            if (!isAuthenticated) {
              this.oidcSecurityService.getAuthorizeUrl().subscribe(url => {
                  const regex = /(.*?)\?(.*)/
                  const regexResult = regex.exec(url);
                  const params = regexResult[2]
                    .split("&")
                    .map(keyValueStr => keyValueStr.split("="))
                    .filter(keyValue => !["nonce", "code_challenge", "code_challenge_method"].includes(keyValue[0]))
                    .map(keyValue => keyValue.join("="))
                    .join("&");
                  const newUrl = `${regexResult[1]}?${params}`
                  navigateExternal(newUrl);
                }
              )
            }
          });
```

peroxid:
I'm in the same boat: the IDP doesn't support PKCE and rejects the code exchange. I can't use another IDP, and I can't use a different one, and the IDP can't be made to support PKCE, for stupid large company reasons. It would be great if this PR could be merged. Maybe add stern warnings to the documentation, or even name the parameter "downgradeSecurityAndDisablePkce" or something like that so that no one will unneccessarily use it.

FabianGosebrink:
Can you, @Kwoin , merge main back into your PR to get the builds green?

damienbod:
Thanks @Kwoin 

@FabianGosebrink I think we merge this once the PR is ready

- default must use PKCE 
- We add this to the docs that it is not recommended

Kwoin:
merge done.

FabianGosebrink (approved):

PR #1491, opened 2022-07-28
fix: RxJS imports to be compatible with RxJS 6

### LINKED ISSUE #1490: [Feature Request]: Go to signup instead of signin
Hey!
Im using the library with Keycloak as my OIDC Identity Provider... And every time that i use the authorize or authorizeWithPopUp it goes to the login page. And its fine, its developed to do that but i'd like to redirect user to the signup page.

I was wondering... If you can add another function method, to allow users to go to the signup page directly, instead of login/signin.

Thanks!!!

timdeschryver:
Move `operators` from `rxjs` to `rxjs/imports`
Closes #1490


damienbod:
Thanks @timdeschryver

Issue #1482, opened 2022-07-06
[Bug]: EventService does not fire events anymore


JohnGalt1717:
**What Version of the library are you using?**
14.1.1

**Describe the bug**
In V13 in your AuthenticationModule code you could run something like this:

		eventService
			.registerForEvents()
			.pipe(filter((notification) => {
				return notification.type == EventTypes.ConfigLoaded || notification.type == EventTypes.ConfigLoadingFailed;
			}))
			.subscribe(() => {
				userService.Init();
			});
			
The pipe/filter never runs nor subscribe ever fires in V14 anymore.

**To Reproduce**
Take any of the examples that uses this code, and upgrade to v14 and run it. It doesn't break into the code.

**Expected behavior**
Should still fire properly.

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser Chrome
- Version Latest

FabianGosebrink:
Have you tried it without the `filter` operator? So does it not fire at all or only when fitering those Events?

JohnGalt1717:
It never fires at all.

Here's the module that does the loading:

```
@NgModule({
	imports: [CommonModule, AuthModule.forRoot({
		config: {
			authority: apiUrl(),
			configId: AUTH_CONFIG_ID,
			startCheckSession: true,
			redirectUrl: `${location.protocol}//${location.host}`,
			postLogoutRedirectUri: LOGOUT_PATH,
			postLoginRoute: DEFAULT_LOGIN_REDIRECT_PATH,
			clientId: CLIENT_ID,
			scope: 'openid profile email',
			responseType: 'code',
			silentRenew: true,
			silentRenewUrl: `${location.protocol}//${location.host}/silent-renew.html`,
			logLevel: environment.production ? LogLevel.Error : LogLevel.Debug,
			autoUserInfo: true,
			forbiddenRoute: FORBIDDEN_PATH,
			autoCleanStateAfterAuthentication: true,
			triggerAuthorizationResultEvent: true,
			maxIdTokenIatOffsetAllowedInSeconds: 600,
			unauthorizedRoute: UNAUTHORIZED_PATH,
			renewTimeBeforeTokenExpiresInSeconds: 20,
			ngswBypass: true,
			renewUserInfoAfterTokenRenew: true,
		}
	})],
	providers: [
		{
			provide: AbstractSecurityStorage,
			useClass: DefaultLocalStorage,
		},
		{
			provide: BaseUserService,
[...]

PR #1494, opened 2022-08-05
Fix throwing config event

FabianGosebrink:
#1482

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/config/config.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/config/config.service.ts
@@ -112,23 +112,30 @@ export class ConfigurationService {
     const usedConfig = this.prepareConfig(passedConfig);
     this.saveConfig(usedConfig);
 
-    const alreadyExistingAuthWellKnownEndpoints = this.storagePersistenceService.read('authWellKnownEndPoints', usedConfig);
-    if (!!alreadyExistingAuthWellKnownEndpoints) {
-      usedConfig.authWellknownEndpoints = alreadyExistingAuthWellKnownEndpoints;
-      this.publicEventsService.fireEvent<OpenIdConfiguration>(EventTypes.ConfigLoaded, usedConfig);
+    const configWithAuthWellKnown = this.addWellKnownEndpoint(usedConfig);
 
-      return of(usedConfig);
-    }
-
-    const passedAuthWellKnownEndpoints = usedConfig.authWellknownEndpoints;
-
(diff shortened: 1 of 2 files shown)


projects/angular-auth-oidc-client/src/lib/config/config.service.ts
```diff
@@ -112,23 +112,30 @@ export class ConfigurationService {
...
+
+  private addWellKnownEndpoint(configuration: OpenIdConfiguration): OpenIdConfiguration {
+    const alreadyExistingAuthWellKnownEndpoints = this.storagePersistenceService.read('authWellKnownEndPoints', configuration);
     if (!!alreadyExistingAuthWellKnownEndpoints) {
-      usedConfig.authWellknownEndpoints = alreadyExistingAuthWellKnownEndpoints;
-      this.publicEventsService.fireEvent<OpenIdConfiguration>(EventTypes.ConfigLoaded, usedConfig);
+      configuration.authWellknownEndpoints = alreadyExistingAuthWellKnownEndpoints;
 
-      return of(usedConfig);
+      return configuration;
```
damienbod, line 127: The return type is changed, will this not effect methods using this?

projects/angular-auth-oidc-client/src/lib/config/config.service.ts
```diff
@@ -112,23 +112,30 @@ export class ConfigurationService {
...
 
-    const alreadyExistingAuthWellKnownEndpoints = this.storagePersistenceService.read('authWellKnownEndPoints', usedConfig);
+    const configWithAuthWellKnown = this.addWellKnownEndpoint(usedConfig);
+
+    this.publicEventsService.fireEvent<OpenIdConfiguration>(EventTypes.ConfigLoaded, configWithAuthWellKnown);
+
+    return of(usedConfig);
+  }
+
+  private addWellKnownEndpoint(configuration: OpenIdConfiguration): OpenIdConfiguration {
```
damienbod, line 122: This name can be improved? It is not just add I think

projects/angular-auth-oidc-client/src/lib/config/config.service.ts
FabianGosebrink: The return type did not change, because this is a new private method. The return type of the public method remains the same.

projects/angular-auth-oidc-client/src/lib/config/config.service.ts
FabianGosebrink: Improved the name

damienbod:
thanks!

Issue #1462, opened 2022-06-20
[Feature Request] Support Observable<OpenIdConfiguration[]> in StsConfigHttpLoader constructor


nwpBerthel:
We are having multiple Server instances and fetching the config from a single Endpoint (per Server), therefore I need to pass the authConfigs from only one async source, but an array of configs nevertheless.

example:
```
export const httpLoaderFactory = (httpClient: HttpClient) => {
  const authConfigs$ = httpClient.get<OpenIdConfiguration[]>(`https://...`);
  return new StsConfigHttpLoader(authConfigs$);
};
```

In an earlier version I could just fetch the configs "by hand" and then initialize the service with the plain array. (`OidcConfigService.withConfigs()`)
Since now I am forced to do this in the module declaration (which is definitely the cleaner way) I can't pass it that way anymore.

Currently only one config from an async source or multiple configs each from another async cource are supported (`constructor(configs$: Observable<OpenIdConfiguration> | Observable<OpenIdConfiguration>[]);`), both of which don't suite my use-case.
Also I can't figure out how to split the one Observable into many (and since the number of entries are dynamic, I can't create Observables by hand and use something like `shareReplay`).
Any help to achieve that split into multiple Observables is much appreciated.

FabianGosebrink:
Hey @nwpBerthel maybe the PR is it. If you want, you can double-check.

nwpBerthel:
Thanks a lot, looks good to me and exactly what I was wishing for.
Also great work not intruducing a breaking change which I though was necessary, but yeah, with more conversions and checks no breaking changes needed :)

FabianGosebrink:
We are doing everything for our users :) Glad you like the change and glad I could help. Thanks for using our lib!

FabianGosebrink:
Released.

PR #1496, opened 2022-08-05
Fabiangosebrink/support observable open id configuration in sts config http loader constructor

FabianGosebrink:


FabianGosebrink:
Fixing #1462

Issue #1339, opened 2021-12-30
[Bug]: DataError: The required JWK member "kty" was missing


ngCorey:
**Critical Notes**
I am working/testing inside of a closed network which prevents me from posting full internal URLs for security purposes.  I've kept as much information intact as possible.

**Environment:**
angular-auth-oidc-client: 13.1.0
Angular CLI: 13.0.0
Node: 16.13.1
NPM: 8.3.0
OS: win32 x64
Browser: Chrome v. 96.0.4664.110

**Bug Description:**
After executing a GET  on /openid-connect-provider/discovery/keys and receiving the  200 response, I am getting the following error:
[Error] 0-{server client id} - DataError: The required JWK member "kty" was missing

**Reproduce**
1. Build project based on the following sample: https://github.com/damienbod/angular-auth-oidc-client/tree/main/projects/sample-code-flow-refresh-tokens
2. Run the app.

**Expected Behavior**
With the "kty" member showing in the GET response, I would expect the module to continue on.

**Screenshots**
The screenshot below is the Chrome Debugger as the login process was occurring.
![Chrome_Debug](https://user-images.githubusercontent.com/48068201/147788216-21fb40af-0d9d-4798-9591-0ae3462959fc.jpg)

The screenshot below is a preview of the response from a GET on https://xxxx.xxxx.com/openid-connect-provider/discovery/keys.
![Chrome_Debug_1](https://user-images.githubusercontent.com/48068201/147788217-5b416ab8-cd55-41fe-9164-ed6211b51e83.jpg)

**Additional context**
I've compared additional the json return for ../discovery/keys from a couple other providers all examples look the same.

damienbod:
Hi @ngCorey 

I need a bit more info, I don't see what is wrong here. Does the JWT URL exist on the well known endpoints and does this match?

In the network tab, does the request to the JWK URL fail? I have not encountered this problem with multiple different  STS implementations and different key types.

Greetings Damien

ngCorey:
The GET to the jwksUri (https://ixxx.xxx.com/openid-connect-provider/discovery/keys) succeeds and there is JSON response.

Here's the complete cycle from loading the page until the error happens based on the :

1.  AuthModule
import { NgModule } from '@angular/core';
import { AuthModule, LogLevel } from 'angular-auth-oidc-client';

@NgModule({
  imports: [
    AuthModule.forRoot({
      config: {
[...]

PR #1517, opened 2022-08-25
Bugfix/unable to extract jwk without kid

erichjsonfosse:
**Fixes**: #1339 
**Tests passing?**: Affirmative

- Added `JwkExtractor`, a service that conforms with [RFC7517](https://www.rfc-editor.org/rfc/rfc7517) when it comes to the `kty`, `kid` and `use` claims.
- Added `JwkWindowCryptoService`, a "sibling" to `JwtWindowCryptoService`
- Added an extra test case for `TokenValidationService.validateSignatureIdToken` to test for a successful result

**Description**:
The `TokenValidationService` now tries to extract keys based on `kty` and `use`, and `kid` if it exists. If no matches are found, it tries to extract keys based on `kty`, and `kid` if it exists (`kty` is required by the spec). If no matches are found, a sensible error is thrown. If matches are found, the first one is chosen, regardless of the number of matches.

Tested `sample-code-flow-refresh-tokens` successfully with https://offeringsolutions-sts.azurewebsites.net when running the project locally.

Let me know if anything is missing or not up to expectations, and I'll update the PR asap :+1:

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/validation/jwk-window-crypto.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/validation/jwk-window-crypto.service.spec.ts
@@ -44,10 +44,10 @@ describe('JwkWindowCryptoService', () => {
     });
   });
 
-  beforeEach(async () => {
+  beforeEach(waitForAsync(() => {
     service = TestBed.inject(JwkWindowCryptoService);
-    await service.importVerificationKey(key3, alg).then((c) => cryptoKey = c);
-  });
+    service.importVerificationKey(key3, alg).then((c) => cryptoKey = c);
+  }));
 
   it('should create', () => {
     expect(service).toBeTruthy();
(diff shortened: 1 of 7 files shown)


andreaslarssen:
@damienbod Could you find the time to check out this PR? It's fixing what stops us from upgrading from Angular 13

projects/angular-auth-oidc-client/src/lib/validation/jwk-window-crypto.service.spec.ts
```diff
@@ -44,9 +44,9 @@ describe('JwkWindowCryptoService', () => {
     });
   });
 
-  beforeEach(() => {
+  beforeEach(async () => {
```
FabianGosebrink, line 47: use waitforasync

erichjsonfosse: @FabianGosebrink Thanks for the feedback! How about now? :smile:

erichjsonfosse:
> Thanks @erichjsonfosse PR LGTM I do some manual testing and release, all goes well, this weekend
> 
> One test is failing

Thanks for the feedback @damienbod ! Managed to reproduce locally now. Working on it.

erichjsonfosse:
@damienbod Both the `test-lib` and `test-lib-ci` runs successfully locally after the latest update. Awaiting feedback :)

FabianGosebrink:
Thanks so much @erichjsonfosse !!!!! ❣️

PR #1526, opened 2022-09-05
Fix checkSession messageHandler configuration

jerousseau:
`messageHandler` need `OpenIdConfiguration` type in first parameter, but `configId: string` was passed in `this.messageHandler.bind`, so `authWellKnownEndPoints` return null value.


Norby125:
@damienbod @FabianGosebrink could we get a new release that contains this change? When is it planned? I'm sorry if I shouldn't have commented here, but if that's the case, can you point me where I can ask for a new release? Thank you.

ccondrup:
@Norby125 this was just released in [14.1.5](https://github.com/damienbod/angular-auth-oidc-client/releases/tag/release_14_1_5)

Issue #1505, opened 2022-08-15
[Bug]: Wrong offset used when calling `validateIdTokenExpNotExpired` from `StateValidationService`


kaep7n:
**What Version of the library are you using?**
1.14.3

**Describe the bug**
After the upgrade to the newest version and Angular 14 we were receiving the following error messages:

```
[DEBUG] Has idToken expired: true --> expires in -2:0-30 , 3:30:38 PM > 3:32:08 PM
-- the times above are local times UTC looks like this: 13:30:38 > 13:32:08
[WARN] authCallback id token expired
[DEBUG] authCallback token(s) invalid
[WARN] authorizedCallback, token(s) validation failed, resetting. Hash: 
[ERROR] Error: authorizedCallback, token(s) validation failed, resetting. Hash:
```

The Id token had an `iat` of `13:30:08` and an `exp` of 1`3:30:38` the current time was `13:30:08` (all times UTC) so the token should be valid. However the validation service stated that the current time was 13:32:08.

**To Reproduce**

Use an Id token expiration below 2 Minutes.

**Expected behavior**
The Id token should be valid, since it is not expired.

**Additional context**

I think this change ([192d97f9228d59e5227bad032c80ab7bf504ea1e](https://github.com/damienbod/angular-auth-oidc-client/commit/192d97f9228d59e5227bad032c80ab7bf504ea1e#diff-90ea606791e3fd3ee70a9f1238f7c2ba9da5b38335a8138ab12ad9aa3e310164)) introduced the problem, it forwards the field `maxIdTokenIatOffsetAllowedInSeconds` from the current `OpenIdConfiguration` to the method `validateIdTokenExpNotExpired`o of the class `TokenValidationService`. 

That causes the service to compare the expiration value of the Id token to Utc Now + 2 Minutes. (If `maxIdTokenIatOffsetAllowedInSeconds` is not set in the configuration.

**Workaround**
Set ``maxIdTokenIatOffsetAllowedInSeconds` to a lower value or use `disableIdTokenValidation = true` ( I don't think thats a good idea)

liamkeaton:
I've had a similar issue and created a PR that should resolve the issue thanks to @kaep7n finding the root cause

Issue #1522, opened 2022-08-31
[Question]: Infinite loop after login when maxIdTokenIatOffsetAllowedInSeconds is configured


walter-psjr:
**What Version of the library are you using?**
14.1.4

**Question**
Hi, I updated an angular 13 project with angular-auth-oidc-client 13 to angular 14 with angular-auth-oidc-client 14.1.4. In the old version I was making use of the maxIdTokenIatOffsetAllowedInSeconds option because many of our users had the clock with the wrong time and everything was working perfectly. But after updating, using the same settings, now the application is in an infinite loop after login, appearing and disappearing the url parameter ?code=... and the following message appears on the console:

**authCallback id token expired
authorizedCallback, token(s) validation failed, resetting. Hash:**

If I remove the maxIdTokenIatOffsetAllowedInSeconds configuration it works again. Could anyone let me know if there was any change in behavior regarding this configuration or if it could be a possible problem?

L1nkss:
Possibly because of my [PR](https://github.com/damienbod/angular-auth-oidc-client/pull/1451/files), but I'm not sure
I had the same problem as you 
Did you try to add `disableIdTokenValidation: true` to your configuration?

walter-psjr:
Hi @L1nkss , 

Yes, with disableIdTokenValidation: true it works, but without it and with maxIdTokenIatOffsetAllowedInSeconds the application enters a loop after the post login redirect. So would it be a bug caused by this pull request?

walter-psjr:
Hi, any clues about this problem? I'm stuck not being able to release an application update because of this. Could anyone tell me what would be the risk of disabling this idtoken validation? Could this expose any risk?

walter-psjr:
The problem has been fixed in version 14.1.5.

Thanks

Issue #1515, opened 2022-08-18
Silent renew process over PKCE seems to use renewTimeBeforeTokenExpiresInSeconds not the way you would expect - possibly a BUG


silviu-sterian:
**What Version of the library are you using?**
Using v 14.1.3

**Question**
How can I trigger the renewal of the token a certain number of seconds before it expires?
The recommended way is setting renewTimeBeforeTokenExpiresInSeconds = xx but this seems to just give an offset to the expiration time.  I wish it would trigger the silent renew xx seconds before token comes back as expired, not expiring the token sooner

In short, I looked at the code in TokenValidationService -> validateAccessTokenNotExpired(..)  and confirmed the logic is offsetting the expiration time but all it does is expire it sooner. So it will come back as expired during this check early (by the number of seconds). 
This does not help closing the small window where token shows as expired and if a user clicks on anything and we're trying to get the token for an api call it will trigger the authentication process - round-trip to the identity provider.

What I was hoping I can get is an early trigger of the silent renewal process - by a number of seconds, while the token is still valid for the number of seconds. Can you help me with an workaround or let me know if you;re considering adding this feature?

**At this point I am pretty sure it is a bug. Ideally we should either change the function calling validateAccessTokenNotExpired(..)  to not pass in the offset and just trigger the process early or give a user of the library more control to trigger the process based on a custom trigger.**

The screenshot attached is representative of the issue. The expiration time should not be negative: 

![image](https://user-images.githubusercontent.com/38025318/185714054-ea53fe83-1a61-4669-8cea-434b9c821a15.png)

silviu-sterian:
![image](https://user-images.githubusercontent.com/38025318/185658482-b0d61737-b76e-4372-9e32-c2b3a08d2ba6.png)

andreaslarssen:
Anyone of the maintainers have a comment on this? I'm suspecting a similar experience that I haven't gotten around to debug yet.

dabbid:
Same for me. Without this option, I cannot get silent renew working, while setting this value to 60s with another library (which is now archived, so my switch to this one) was working well.

andreaslarssen:
@damienbod Can you confirm that this is a bug?

damienbod:
@andreaslarssen  I need to look at this first and reproduce to verify

Greetings Damien

silviu-sterian:
Thank you for looking into this @damienbod
Let me know if you need any info, it should be straight forward to reproduce.

[...]

PR #1529, opened 2022-09-06
fix: Use correct offset in ID token expiry check

liamkeaton:
When using a large maxIdTokenIatOffsetAllowedInSeconds the ID token is
always marked as expired. Issue #1505 found this was due to the wrong
offset being passed to the validateIdTokenExpNotExpired method. This
PR uses the renewTimeBeforeTokenExpiresInSeconds which is used in the
auth-state.service.ts hasIdTokenExpiredAndRenewCheckIsEnabled method for
a similar purpose.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/validation/state-validation.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/validation/state-validation.service.ts
@@ -58,6 +58,7 @@ export class StateValidationService {
         disableIatOffsetValidation,
         ignoreNonceAfterRefresh,
         disableIdTokenValidation,
+        renewTimeBeforeTokenExpiresInSeconds,
       } = configuration;
 
       toReturn.idToken = callbackContext.authResult.id_token;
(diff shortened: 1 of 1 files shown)


walter-psjr:
I think this has to do with an issue i am facing in my application after upgrading to angular and angular-auth-oidc-client 14 - #1522

damienbod:
Thanks @liamkeaton

walter-psjr:
When can we have a new version with this fix @damienbod?

andreaslarssen:
:point_up:

andreaslarssen:
Seems like this also could be the root of #1515

silviu-sterian:
> When can we have a new version with this fix @damienbod?

@andreaslarssen  this unfortunately does not fix my issue.  This issue (reported above) has to do with renewTimeBeforeTokenExpiresInSeconds property and it is not solved by this release

walter-psjr:
It worked for me with version 14.1.5

PR #1533, opened 2022-09-13
Fix bug in logoffAndRevokeTokens() which was not revoking the access token correctly

benp25:
See full issue description in #1534.

This PR is to resolve the issue in the V13 branch.


damienbod:
Not sure if we release this, we usually only fix the latest version due to time

Thanks for the PR

PR #1554, opened 2022-10-06
ci: detect RxJS operators from 'rxjs'

timdeschryver:
We've had this come up before that we've imported rxjs operators from the 'rxjs' namespace. Doing this made it incompatible to projects that are RxJS v6.

A lint rule detects these so we can prevent those "issues" from happening.

I suggested that this would be an ESLint rule within the RxJS Linter, but we can already implement this rule by using the [no-restricted-imports ESLint rule](https://eslint.org/docs/latest/rules/no-restricted-imports).

This PR configures the `no-restricted-imports` rule to throw an error when operators are imported from the `rxjs` namespace.


timdeschryver:
You can see that it works 🤪

projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
-import { delay, of, throwError } from 'rxjs';
+import { of, throwError } from 'rxjs';
+import { delay } from 'rxjs/operators';
```
timdeschryver, line 3: It seems like adding this import also re-formatted the file.
If needed I can revert this.

damienbod: tip top, thanks

Issue #1599, opened 2022-11-22
[Feature Request]: Provide Standalone APIs (A15)


FabianGosebrink:

timdeschryver:
👋 Can I help with anything here?
I'm thinking of a functional interceptor, but perhaps there are more opportunities here.

FabianGosebrink:
Of course, you can help here, this is always appreciated. I have to dive into this a bit, but I am open for every proposal. Would also love to have a sample app to show the users how to use this lib with standalone APIs.

timdeschryver:
@FabianGosebrink do you (or someone else) are working on an standalone example?
If that's not the case, I should have some time later to look into this if that's OK.

FabianGosebrink:
Hey Tim, no I do not, and I don't think anybody else does. Yes, that would be nice. I think to get standalone working, we have to provide the providers array of the lib https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/auth.module.ts#L84-L145 as a method which gets exported and then used by standalone components. Not sure tho.

FabianGosebrink:
Yes, we are planning to move and provide full standalone support. I am short on time, think it will be July, then I have time to do it. But it will come.

FabianGosebrink:
We merged several PRs for standalone APIs, @timdeschryver did a great job there and with the next release this is gonna be released as well. Thanks!

PR #1610, opened 2022-11-26
feat: add functional interceptor

timdeschryver:
This PR adds a functional auth interceptor, so it can be used with the new standalone API.

Part of #1599

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/interceptor/auth.interceptor.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/interceptor/auth.interceptor.spec.ts
@@ -1,11 +1,11 @@
-import { HttpClient, HTTP_INTERCEPTORS } from '@angular/common/http';
-import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
+import { HttpClient, HTTP_INTERCEPTORS, provideHttpClient, withInterceptors } from '@angular/common/http';
+import { HttpClientTestingModule, HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
 import { TestBed, waitForAsync } from '@angular/core/testing';
 import { mockClass } from '../../test/auto-mock';
 import { AuthStateService } from '../auth-state/auth-state.service';
 import { ConfigurationService } from '../config/config.service';
 import { LoggerService } from '../logging/logger.service';
-import { AuthInterceptor } from './auth.interceptor';
+import { authInterceptor, AuthInterceptor } from './auth.interceptor';
 import { ClosestMatchingRouteService } from './closest-matching-route.service';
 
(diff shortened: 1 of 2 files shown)


projects/angular-auth-oidc-client/src/lib/interceptor/auth.interceptor.ts
```diff
@@ -16,46 +16,68 @@ export class AuthInterceptor implements HttpInterceptor {
...
+      loggerService: inject(LoggerService),
+    });
+  };
+}
 
-    const { matchingConfig, matchingRoute } = this.closestMatchingRouteService.getConfigIdForClosestMatchingRoute(
-      req.url,
-      allConfigurations
-    );
+function interceptRequest(
```
timdeschryver, line 39: I don't know if there's a better way, but I extracted the logic into this `interceptRequest` method so it can be reused within the class interceptor and the functional interceptor.

projects/angular-auth-oidc-client/src/lib/interceptor/auth.interceptor.spec.ts
```diff
@@ -15,42 +15,78 @@ describe(`AuthHttpInterceptor`, () => {
...
+      httpTestingController.verify();
+    });
+
+    runTests();
   });
 
-  it(
-    'should add an Authorization header when route matches and token is present',
-    waitForAsync(() => {
+  function runTests() {
```
timdeschryver, line 88: The same tests are run for the class and functional interceptors.
The only difference is the setup in the `beforeEach` block.

damienbod:
LGTM, would love @FabianGosebrink opinion.

projects/angular-auth-oidc-client/src/lib/interceptor/auth.interceptor.ts
FabianGosebrink: Yes, that looks fine to me. Thanks. I think I would have done it the same way.

FabianGosebrink (approved):

FabianGosebrink:
LGTM as well, thanks @timdeschryver . 

This is the interceptor, we have to support the providers in a method as well so that users can use this lib without modules. But this can be another PR.

The PR fails with A14 with a *.d.ts method. Any ideas why this is happening? Thanks :)

timdeschryver:
That's probably because it's a v15 type.
Let me see what I can do about it tomorrow :) 

Thanks for the reviews!

FabianGosebrink:
Thank YOU for your work on this! We have to support A14 as well 😑

timdeschryver:
Pfew, it took a bit longer than expected 😅

FabianGosebrink:
You are incredible. Thank you so much!

PR #1639, opened 2022-12-20
docs: add standalone example

timdeschryver:
This PR adds a standalone example, which is based on the "auto-login-flow" example.
The example can be used/updated for #1599.
We should also still need to update the docs to include this example when the example is using the standalone API's.

oidc


projects/sample-code-flow-standalone/src/main.ts
```diff
@@ -0,0 +1,35 @@
...
+import { AppComponent } from './app/app.component';
+import {HomeComponent} from "./app/home/home.component";
+import {ProtectedComponent} from "./app/protected/protected.component";
+import {ForbiddenComponent} from "./app/forbidden/forbidden.component";
+import {UnauthorizedComponent} from "./app/unauthorized/unauthorized.component";
+import {AuthConfigModule} from "./app/auth-config.module";
+
+bootstrapApplication(AppComponent, {
+  providers: [
+    importProvidersFrom(AuthConfigModule),
```
timdeschryver, line 18: `importProvidersFrom` needs to be replaced later with the standalone API

timdeschryver: Can be replaced if #1643 lands 😀

FabianGosebrink:
Hey Tim, thanks, we really appreciate your work. We should first merge #1643, right?

PR #1643, opened 2022-12-23
feat: add provideAuth for standalone applications

timdeschryver:
This PR extracts the config into the `_provideAuth` method, which in its turn is used in the existing `AuthModule` and the new `provideAuth` method (this is the best name that I can think of, feel free to suggest another method name).

The config is also extracted into the `auth-config` file so it can be reused.

This solution is based on how we did it with the NgRx APIs. 

Part of #1599


damienbod:
This looks really great. @FabianGosebrink @timdeschryver Can we merge both of these PRs? 

I will use the docs from the the PR to test.

Anything missing?

Thanks for the 2 PRs

timdeschryver:
Happy new year @damienbod and @FabianGosebrink , my best wishes 🎆

To answer https://github.com/damienbod/angular-auth-oidc-client/pull/1639#issuecomment-1364947968, yes I think we should first merge this PR and then update the other one (sorry for the confusion, but I had some time left for this PR). 

I don't know how to make the CI pipeline happy though. To create the standalone provider we're using the `makeEnvironmentProviders` method from Angular, which is only available in v15 AFAIK. Previous errors were fixed by re-declaring the type/interface on our end, but since this is a method we can't just reimplement it on our side.

FabianGosebrink:
Hey @timdeschryver , thanks for your wishes. Happy new year and all the best for you, too. 

This would mean we would drop support for A14, right?

Thanks

timdeschryver:
Yea, I can't think of a way to support A14 (see latest commits where I tried some things).
But, for me personally, this PR can wait until we got more breaking changes though, or until more members from the community are asking for this feature.

FabianGosebrink:
Hey @timdeschryver , thanks for this PR. I would say that we merge this one once A16 is out. Then we can drop A14 support IMHO. What do you think?

timdeschryver:
Sounds great @FabianGosebrink !
I will check from time to time to rebase this PR .

damienbod:
@timdeschryver Thanks a million for these 2 PRs

@timdeschryver @FabianGosebrink 

I think now is the right time to complete these PRs and add this. Good?

Greetings Damien

timdeschryver:
If we drop support for A14, we can also clean up the "this type can be dropped when Angular 14 support" comments.

timdeschryver:
@FabianGosebrink if it's OK for you I can also create a PR to drop support for A14.

PR #1665, opened 2023-01-30
adding provided in root to services

FabianGosebrink:


FabianGosebrink:
@timdeschryver do you see any bugs when we provide every service with `providedin: root`? What do you think of this PR?

timdeschryver:
@FabianGosebrink When I was adding the standalone functions, I was also think about this.
I think this can be done, we did something similar while introducing the standalone functions in NgRx.

FabianGosebrink:
Thanks for your quick reply @timdeschryver :)

Issue #1674, opened 2023-02-04
[Question]: Move Http Interceptor into its own page


timdeschryver:
**What Version of the library are you using?**
N/A

**Question**
In most applications we need to add an authorization header to the requests that the application makes. This is done with HTTP Interceptors, and this library already provides the `AuthInterceptor` interceptor.
But, this is a bit hidden in the current documentation. The `AuthInterceptor` interceptor is currently added to the "Using Access Tokens" page.
Because it's such a frequent thing to look up, I think it would be better to move the current content in a dedicated "Interceptor" page.

I opened this question because I saw new developers writing their own interceptor because they didn't find/search the docs.

FabianGosebrink:
Hey, if you want to review, feel free :)

PR #1680, opened 2023-02-12
[Question]: Move Http Interceptor into its own page #1674

FabianGosebrink:
#1674


FabianGosebrink:
@timdeschryver would you like to review

timdeschryver (approved):
It's great, thanks!

A few nits (I can't comment in the file because the lines are untouched):

- `You can configure the routes you want to send a token with in the configuration` should be `You can configure the routes you want to send a token within the configuration`

- `The lib provides an own interceptor implementation which` should be `The lib provides its own interceptor implementation which`

FabianGosebrink:
Pushed. With your OK I'll merge.

timdeschryver:
Hereby I give you my formal OK 😁

Issue #1746, opened 2023-05-02
[Bug]: Silent renew not starting on sample-code-flow-auto-login-all-routes


bluebaroncanada:
### Version

current main branch

### Please provide a link to a minimal reproduction of the bug

https://github.com/damienbod/angular-auth-oidc-client/tree/main/projects/sample-code-flow-auto-login-all-routes

### Please provide the exception or error you saw

The silent renew process does not start after page refresh.

### Steps to reproduce the behavior

```true
git checkout https://github.com/damienbod/angular-auth-oidc-client.git
cd angular-auth-oidc-client
npm run build
npm run start-sample-code-flow-auto-login-all-routes
browse to https://localhost:4204
login
refresh the page
```

### A clear and concise description of what you expected to happen.
The silent-renew service should start after page refresh if authenticated.

bluebaroncanada:
The issue is entirely my fault some how.  I don't know exactly what I did wrong.  I completely reinstalled a few times, but I was somehow missing the signs that it was actually working now.  Sorry for the trouble.  You guys seem exhausted.  Thank you for all your great work.

bluebaroncanada:
Actually, I have further confirmed this issue is happening with your example.

When you first login:
![image](https://user-images.githubusercontent.com/5856849/236003677-349a284c-b8f2-4259-8940-b3c5803adab0.png)

Click on protected, refresh, and the silent renew doesn't start.

![image](https://user-images.githubusercontent.com/5856849/236004052-03761f84-df9c-4870-a7d1-ad5910ebeecd.png)

This pretty much makes the library unusable, but what I don't understand is how I'm the first to point this out.  I've gone through several bug reports and haven't found this exact issue which is why I have so much self-doubt about this issue.

PR #1750, opened 2023-05-05
Issue 1746: Silent renew not starting after refresh

bluebaroncanada:
I deep-dived the code for several hours and I think this is the way you would want this to be fixed.  You have your own CheckAuth function in the guard and I checked all code paths that lead to the `PeriodicallyTokenCheckService.startTokenValidationPeriodically` function and I think there's nothing that applies.

https://github.com/damienbod/angular-auth-oidc-client/issues/1746


bluebaroncanada:
I was unable to confirm the test works.  It may not work.  I'm getting an error with all tests in that file.  I got it to the point where I also receive that error.

Issue #1708, opened 2023-03-03
[Bug]: isAuthenticated$ and isAuthenticated() return different values when logged in and reloading the page


heapifyman:
### Version

15.0.3

### Please provide a link to a minimal reproduction of the bug

https://github.com/damienbod/angular-auth-oidc-client/tree/main/projects/sample-code-flow-auto-login-all-routes

### Please provide the exception or error you saw

```true
`authenticated:  false home.component.ts:19:14`
```

### Steps to reproduce the behavior

```true
1. Run sample app [Code flow PKCE auto-login for all routes](https://github.com/damienbod/angular-auth-oidc-client/tree/main/projects/sample-code-flow-auto-login-all-routes)
2. Login and get redirect to HomeComponent - it will display "Is Authenticated: true" and your user data.
3. Press F5 to reload the page - it will display:

> Is Authenticated: false

> {
>   "userData": null,
>   "allUserData": []
> }

And the navigation bar at the top will return to the "not logged in state", only displaying "Route to 'protected' Route & Auto Login".

However, clicking the "Route to 'protected' Route & Auto Login" link in the navigation bar still works. You can access the "/protected" route and the page displays

> protected works!

All the other routes that require authentication like "/customers" or "/forbidden" also still work.
```

### A clear and concise description of what you expected to happen.

```true
After pressing F5 the HomeComponent should still display the logged in user's data and the navigation bar should remain in the "user is logged in" state, showing all the links.
```

### Additional context

It appears that the `isAuthenticated$` and `userData$` observables do not emit the correct value when the page is reloaded.

`isAuthenticated$` will emit `false` even though the user is still logged in and the other observable `isAuthenticated()` still returns `true`. Can be tested by changing the code of `NavigationComponent.ngOnInit` to this:

```
  ngOnInit() {
    this.oidcSecurityService.isAuthenticated$
      .pipe(
        tap(({ isAuthenticated }) => console.log('[NavigationComponent] isAuthenticated$: ', isAuthenticated)),
        concatMap((_) => this.oidcSecurityService.isAuthenticated())
      )
      .subscribe((isAuthenticated) => {
        console.log('[NavigationComponent] authenticated: ', isAuthenticated);
        this.isAuthenticated = isAuthenticated;
      });
  }
```

It will print to the browser console:

> [NavigationComponent] isAuthenticated$:  false
> [NavigationComponent] authenticated:  true

[...]

PR #1752, opened 2023-05-05
Issue 1746: Silent renew not starting after refresh

bluebaroncanada:
Fixes for https://github.com/damienbod/angular-auth-oidc-client/issues/1746

Issue:
Silent renew is not starting after page refresh.  

Solution:
In `AutoLoginPartialRoutesGuard.checkAuth`, if `isAuthenticated` call `this.periodicallyTokenCheckService.startTokenValidationPeriodically`

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login-partial-routes.guard.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login-partial-routes.guard.spec.ts
@@ -10,6 +10,8 @@ import { LoginService } from '../login/login.service';
 import { StoragePersistenceService } from '../storage/storage-persistence.service';
 import { AutoLoginPartialRoutesGuard } from './auto-login-partial-routes.guard';
 import { AutoLoginService } from './auto-login.service';
+import { PeriodicallyTokenCheckService } from "../callback/periodically-token-check.service";
+import { HttpClientTestingModule } from "@angular/common/http/testing";
 
 describe(`AutoLoginPartialRoutesGuard`, () => {
   let autoLoginPartialRoutesGuard: AutoLoginPartialRoutesGuard;
@@ -18,11 +20,12 @@ describe(`AutoLoginPartialRoutesGuard`, () => {
   let storagePersistenceService: StoragePersistenceService;
   let configurationService: ConfigurationService;
   let autoLoginService: AutoLoginService;
+  let periodicallyTokenCheckService: PeriodicallyTokenCheckService;
   let router: Router;
 
   beforeEach(() => {
     TestBed.configureTestingModule({
-      imports: [RouterTestingModule],
+      imports: [RouterTestingModule, HttpClientTestingModule],
       providers: [
         AutoLoginService,
         { provide: AuthStateService, useClass: mockClass(AuthStateService) },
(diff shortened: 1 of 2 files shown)


bluebaroncanada:
Note: From what I'm seeing, this might also affect implicit flow.  I did not test that.
Sorry for not leaving a trailing comma.  That was impolite.

FabianGosebrink:
This would mean that on every refresh on a page with the guard, the periodically token check would be started. Is that correct?

bluebaroncanada:
If it meets the checks inside of `periodicallyTokenCheckService.startTokenValidationPeriodically`.  I actually traced all calls to that function, and I think no other lines of code relate in those issues, and the function itself protects against scenarios where this shouldn't start.

bluebaroncanada:
Also, if this protects all routes except for the Callback route, then it protects all routes because the Callback route calls `OidcSecurityService.CheckAuth` which will eventually call `PeriodicallyTokenCheckService.startTokenValidationPeriodically`.

FabianGosebrink:
Yeah, the sense was, that the `checkAuth()` sets up that call. You can also call that method in your AppModule constructor. Then the issue should be gone, no?

bluebaroncanada:
I cannot because there is no export for `PeriodicallyTokenCheckService`.

bluebaroncanada:
This issue affects the sample out of the box.

bluebaroncanada:
I also believe there are related tickets from other users.

FabianGosebrink:
Talking about the `checkAuth()` method. The use should not call any internal methods to start something by its own. The library should handle that by calling `checkAuth()`. We have to look how we can fix this bug.

bluebaroncanada:
I actually do somewhat agree with your premise, but I think this falls in line with the typical functionality of this guard because it implements its own `CheckAuth` function which also calls other things that should also be abstracted.  I think maybe we'd have to implement an single `CheckAuth` function that is resilient to these code changes.

bluebaroncanada:
If you exported `PeriodicallyTokenCheckService` so I could call it from my `app.component`, that would be a satisfactory solution for me, but I feel it violates the same constraints you outline.

bluebaroncanada:
If you reject this commit and choose a different solution, I would highly recommend considering this bug https://github.com/damienbod/angular-auth-oidc-client/issues/1708 for resolution at the same time as they're probably similar and caused by implementing guard-side `CheckAuth` functions.

FabianGosebrink:
You can edit your replies btw 🙂 We have to fix this bug. Just thinking about the correct way of doing it...

bluebaroncanada:
That bug has been open since March, though, and this issue is a killer, so we were actually a bit worried that this might be abandoned.  The other issue, we can get around.  This one we cannot, and we have customers that are complaining that they're timing out in the middle of logging notes in our application.  We cranked the token expiration to 10hrs for now, but it's still going to be an issue.

bluebaroncanada:
I actually asked my boss to offer a bounty for this issue, but his boss won't allow it.

bluebaroncanada:
I spent 20 hours of my own time after work to hunt down and resolve this issue off company time.

FabianGosebrink:
Why are you writing "dont read this" and write it anyway?

You can fix this by calling

```
export class AppComponent {
  title = 'sample-code-flow-auto-login-all-routes';

  constructor(public oidcSecurityService: OidcSecurityService) {
    this.oidcSecurityService.checkAuth().subscribe(/* ... */);
  }
}
```

in your app component.

bluebaroncanada:
giphy (13)

Sweet!  Thank you so much!  That fixes BOTH issues.  I couldn't be happier!

PR #1774, opened 2023-06-10
feat: add functional guards

### LINKED ISSUE #1758: [Refactoring]: CanActivate is Deprecated
Since Angular version 15 the interface [CanActivate](https://angular.io/api/router/CanActivate) is deprecated. In the future this needs to be refactored.

timdeschryver:
Closes #1758

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login-all-routes.guard.ts
+++ b/projects/angular-auth-oidc-client/src/lib/auto-login/auto-login-all-routes.guard.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@angular/core';
+import { Injectable, inject } from '@angular/core';
 import {
   ActivatedRouteSnapshot,
   Router,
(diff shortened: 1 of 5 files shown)


projects/angular-auth-oidc-client/src/lib/auto-login/auto-login-all-routes.guard.ts
```diff
@@ -17,39 +17,68 @@ export class AutoLoginAllRoutesGuard  {
...
+  const loginService = inject(LoginService);
+  const autoLoginService = inject(AutoLoginService);
+  const router = inject(Router);
+
+  const url = router.getCurrentNavigation()?.extractedUrl.toString().substring(1) ?? '';
+
+  return checkAuth(url, configurationService, checkAuthService, autoLoginService, loginService);
+}
+
+// export function autoLoginAllRoutesGuard(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
```
timdeschryver, line 47: Originally I wanted to create a `ActivateGuard` and a `MatchGuard`, but I don't think this is necessary and it also caused some problems to test this guard (because it requires the `route` and `state` arguments).
This applies to both guards.

Because I spent some time on this I left this here in comments to discuss it before removing it or taking a closer look at the differences/tests.

FabianGosebrink: That sounds good. Thanks!

Issue #1737, opened 2023-04-18
In 15.0.4 authorizeWithPopUp acts totally different then in 15.0.3


majkers:
### Version

15.0.4

### Please provide the exception or error you saw

```true
In 15.0.3 version of library when I used authorizeWithPopUp method my application opened a popUp and after a successful login it closed it. It was ideal for me beacuse I am using two configuration and need to log in to one of them when being logged to other without any redirections. I had a problem with logout in popup but I managed to do that https://github.com/damienbod/angular-auth-oidc-client/issues/1727 But from 15.0.4 the behaviour of authorizeWithPopUp changed. Id doesn't close the popup but statys in it and acts more or less like a new window with the same application. And this is not the behavior I expect. Can You please add a method that acted like authorizeWithPopUp  in 15.0.3 version?
```

majkers:
I think there is an error in logic. Previously there was or between conditions and now it is and so ofr me isCurrentlyInPopup always returs false;

majkers:
One more clue

![image](https://user-images.githubusercontent.com/21099773/232762896-083e2fbc-d86b-40f4-8535-1ab9d9f6f859.png)

adrian-heath:
We are also currently experiencing the same issue with authorizeWithPopPp never completing and emitting from the Observable it returns.  Is there any update on a fix to this issue?

I t is currently holding us back from upgrading to the latest version.

FabianGosebrink:
Wasn't that fixed with the PR merged above?

adrian-heath:
Ah maybe.  I'll try it.  For some reason 'npm outdated' was still showing 15.0.4 as the latest version for me.

Sorry for any confusion

FabianGosebrink:
Easy, no problem. Let us know if you have any issues still :)

adrian-heath:
@FabianGosebrink I tried v15.0.5 this morning and can unfortunately confirm that authorizeWithPopup is never emitting it's observable in 15.0.4 or 15.0.5.  15.0.3 works fine.

We are invoking the method from a popup dialog and the subscribe that closes the dialog no longer fires.  Neither does it error either, but the oAuth dialog is opening correctly and closing once the user enters their login details

    this._oidcSecurityService.authorizeWithPopUp(authOptionsOrNull, popupOptions).subscribe({
      next: ({ isAuthenticated, userData, accessToken, errorMessage }) => {
        this._dialogRef.close(isAuthenticated);

[...]

PR #1776, opened 2023-06-13
moved the setting of popupauth in storage before opening the popup wi…

adrian-heath:
moved the setting of popupauth in storage before opening the popup window due to timing issues meaning the popup didn't see the value

Also added a couple of extra loggin outputs that helped with debugging.


FabianGosebrink:
Fixing #1737

PR #1783, opened 2023-06-19
Release 16 0 0

damienbod:


timdeschryver (approved):
Looks good to me 🚀🎉

Do we need a migration guide for this change?
I think the only breaking change is that angular 14 is not supported.

timdeschryver:
@damienbod I think I should be able to work on the guide this week.
What I had in mind:

- mention we dropped support for angular 14
- add a note about the deprecation of the route guard
- compare the module approach with the standalone/functional approach

Am I missing something here?

FabianGosebrink:
Maybe we can add to the deprecation of the guard that the other `...Partial...` should be used now if you want to secure your complete application and do a link to the sample. Other than that, I think the list is complete. Thanks, @timdeschryver !

PR #1784, opened 2023-06-19
Remove iframe renew, not used in standalone example

damienbod:


timdeschryver (approved):
My bad, good catch 👍
FYI the HTTP interceptors also are not used, but I added these to check if it compiles.

PR #1795, opened 2023-07-01
docs: sort migrations from new to old

timdeschryver:
Order the migrations from new to old, previously this was from old to new.
This makes more sense to me because we're probably interested in the latest first. (this is also how angular does it)

Because auto-format is enabled prettier also formatted the older migrations.

image


damienbod:
LGTM

PR #1800, opened 2023-07-08
feat: support ng-add for standalone applications

### LINKED ISSUE #1798: [Bug]: ng add angular-auth-oidc-client broken on Angular 16 standalone component project
### Version

16.0.0

_No response_

```true
After running ng add angular-auth-oidc-client on an Angular 16 app with standalone components the default configuration is not created because src/app/app.module.ts does not exist.
```

```true
Run ng add angular-auth-oidc-client on an Angular 16 app with standalone components
```

```true
After running ng add angular-auth-oidc-client on an Angular 16 app with standalone components I expect the configuration to work and not report errors.
```

### Additional context

Here is a screenshot of running `ng add angular-auth-oidc-client` on an Angular 16 app with standalone components. The configuration is not completed because `src/app/app.module.ts` does not exist because the Angular 16 app is using standalone components.

image

timdeschryver:
I need to do some more tests, but I already created the PR for initial feedback and to let you know that I've started this.

Closes #1798 

A first test seems like the default migration works for standalone applications.
Here's an overview of the changes in an empty application:

image

image

image

And it still works for a module based application:

image
image


projects/angular-auth-oidc-client/src/lib/index.ts
```diff
@@ -1,5 +1,6 @@
 // Public classes.
 
+export { PassedInitialConfig } from './auth-config';
```
timdeschryver, line 3: Needed to be used to configure a standalone application in a type-safe way.

projects/schematics/src/ng-add/actions/configs.ts
```diff
@@ -24,7 +24,7 @@ const IFRAME_SILENT_RENEW = `{
 
 const AZURE_AD_SILENT_RENEW = `{
             authority: 'https://login.microsoftonline.com/<authorityUrlOrTenantId>/v2.0',
-            authWellknownEndpoint: 'https://login.microsoftonline.com/common/v2.0',
+            authWellknownEndpointUrl: 'https://login.microsoftonline.com/common/v2.0',
```
timdeschryver, line 27: While testing the migration this gave a compile error, so I renamed this property to match the type.

PR #1812, opened 2023-07-27
Check if savedRouteForRedirect is null

jerousseau:
If the default route redirect to a protected route the saved route for redirect is an empty string ("") so after login the user get stuck in callback page without redirect, if we check for null instead the user get redirected to the root


FabianGosebrink:
Thanks! LGTM.

PR #1821, opened 2023-08-16
Added useCustomAuth0Domain flag

fredhair:
Since Auth0 does not publish it's end_session_url in it's discovery doc special handling is required for logging out.
This special handling of Auth0 logout is not triggered for custom domains since this special handling is determined by whether the authority ends with 'auth0.com'

openid-configuration now contains an optional flag to allow custom domains in auth0 to function as normal.
A small change in the url service checks if the authority ends with auth0.com OR the useCustomAuth0Domain is set.

This is probably quite a niche issue but it has been a real headache for what seems like a simple fix.

(This is my first time ever contributing to any project so apologies in advance if I've missed anything / done anything incorrectly, please do help me to learn I would be very grateful!)

### DIFF
--- a/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
+++ b/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
@@ -527,3 +527,10 @@ Default = _false_
 Disable validation for id_token. This is not recommended! You should always validate the id_token if returned.
 
 Default = _false_
+
+### `useCustomAuth0Domain`
+
+- Type: `boolean`
+- Required: `false`
+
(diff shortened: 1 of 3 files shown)


FabianGosebrink:
Thanks for the PR. It would be nice we would extend also the docs with this change. Then it LGTM.

fredhair:
> Thanks for the PR. It would be nice we would extend also the docs with this change. Then it LGTM.

Yes quite agreed, I will take a look at the docs :) thanks for the feedback.

FabianGosebrink:
Haha, nothing to thank for. Thank YOU for the PR!

fredhair:
Configuration doc updated to include a brief description of the useCustomAuth0Domain flag!

PR #1825, opened 2023-08-22
fix problem in logoff when urlHandler option is present

### LINKED ISSUE #1824: [Bug]: Problem when using the urlHandler option in the loggoff method
### Version

15.0.0

### Problem description

When I do a call to the `oidcSecurityService.loggoff` method using the urlHandler method, the log out is not performed and the page is just reloaded. If I remove the `urlHandler` parameter it starts working.

I suspect the problem was introduced in this commit (https://github.com/damienbod/angular-auth-oidc-client/commit/885c841906c2aa3831c3d01a37cc2b348e56535e#diff-58d43ce944501e6a8a7ef737e515754414433f65f03c808d4e1436f89d2529b6R53) but I cannot completely understand why.

In 14.1.5 it works as expected.

### Steps to reproduce the behaviour

Do a `logoff` using `urlHandler` parameter:

```typescript
export class LogoutComponent implements OnInit {
	constructor(private oidcSecurityService: OidcSecurityService) {}

	async ngOnInit(): Promise<void> {
		await firstValueFrom(this.oidcSecurityService.logoff(undefined, {
			urlHandler: (url) => {
				document.location.href = url;
			},
		}));
	}
}
```

_No response_

### Additional context

_No response_

andsouto:
Fixes #1824

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/logoff-revoke/logoff-revocation.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/logoff-revoke/logoff-revocation.service.spec.ts
@@ -8,8 +8,8 @@ import { ResetAuthDataService } from '../flows/reset-auth-data.service';
 import { CheckSessionService } from '../iframe/check-session.service';
 import { LoggerService } from '../logging/logger.service';
 import { StoragePersistenceService } from '../storage/storage-persistence.service';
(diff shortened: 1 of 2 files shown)


FabianGosebrink:
LGTM Thanks!!!

PR #1848, opened 2023-10-02
moving to strict mode

FabianGosebrink:


projects/angular-auth-oidc-client/src/lib/auth-state/auth-state.service.ts
```diff
@@ -78,37 +78,55 @@ export class AuthStateService {
...
+      return '';
     }
 
     const token = this.storagePersistenceService.getAccessToken(configuration);
 
     return this.decodeURIComponentSafely(token);
   }
 
-  getIdToken(configuration: OpenIdConfiguration): string {
+  getIdToken(configuration: OpenIdConfiguration | null): string {
```
vasylchyzh, line 95: getTokenId

projects/angular-auth-oidc-client/src/lib/auth-state/check-auth.service.spec.ts
```diff
@@ -412,9 +439,10 @@ describe('CheckAuthService', () => {
         { configId: 'configId1', authority: 'some-authority' },
       ];
 
-      spyOn(callBackService, 'handleCallbackAndFireEvents').and.returnValue(
-        of(null)
+      spyOn(currentUrlService, 'getCurrentUrl').and.returnValue(
+        'http://localhost:4200'
```
vasylchyzh, line 443: mb it's time to extract it to a separate variable.

projects/angular-auth-oidc-client/src/lib/callback/periodically-token-check.service.spec.ts
```diff
@@ -321,15 +321,15 @@ describe('PeriodicallyTokenCheckService', () => {
 
       tick(1000);
```
vasylchyzh, line 322: DELAY_1000

projects/angular-auth-oidc-client/src/lib/config/default-config.ts
```diff
@@ -4,7 +4,7 @@ import { OpenIdConfiguration } from './openid-configuration';
 export const DEFAULT_CONFIG: OpenIdConfiguration = {
   authority: 'https://please_set',
   authWellknownEndpointUrl: '',
-  authWellknownEndpoints: null,
+  authWellknownEndpoints: undefined,
```
vasylchyzh, line 7: could be omitted

projects/angular-auth-oidc-client/src/lib/config/validation/config-validation.service.ts
```diff
@@ -23,6 +23,10 @@ export class ConfigValidationService {
     passedConfigs: OpenIdConfiguration[],
     allRulesToUse: any[]
   ): boolean {
+    if (passedConfigs.length === 0) {
```
vasylchyzh, line 26: !passedConfigs.length

projects/angular-auth-oidc-client/src/lib/flows/callback-handling/user-callback-handler.service.ts
```diff
@@ -110,9 +114,13 @@ export class UserCallbackHandlerService {
   }
 
   private publishUnauthenticatedState(
-    stateValidationResult: StateValidationResult,
+    stateValidationResult: StateValidationResult | null,
     isRenewProcess: boolean
   ): void {
+    if (!stateValidationResult) {
```
vasylchyzh, line 120: if (!stateValidationResult) return;

projects/angular-auth-oidc-client/src/lib/flows/flows-data.service.ts
```diff
@@ -121,7 +129,11 @@ export class FlowsDataService {
       return false;
     }
 
-    const timeOutInMilliseconds = silentRenewTimeoutInSeconds * 1000;
+    if (storageObject.state === 'not-running') {
```
vasylchyzh, line 132: enum

projects/angular-auth-oidc-client/src/lib/iframe/check-session.service.ts
```diff
@@ -49,17 +48,21 @@ export class CheckSessionService implements OnDestroy {
 
   ngOnDestroy(): void {
     this.stop();
-    this.document.defaultView.removeEventListener(
-      'message',
-      this.iframeMessageEventListener,
-      false
-    );
+    const windowAsDefaultView = this.document.defaultView;
```
vasylchyzh, line 51: better to consider that we have a deal with CSR.
Otherwise, we always have to validate and check whether .Window is defined or not. 
I suggest you provide a dedicated token or forbid access here and validate in the initial state of the lib.

FabianGosebrink:
Ready to review, test and possibly merge @damienbod @timdeschryver  :)

timdeschryver:
> Ready to review, test and possibly merge @damienbod @timdeschryver :)

I got some time later this week, I'll try to review this PR.

FabianGosebrink:
I don't think it is possible to review all files. @damienbod will test, I also did run a few test, worked fine so far.

timdeschryver:
Yea, I'll review some of the files within the library source, and also do a few tests.
I should also be able to build it locally and use it in our application to see if there are surprises there.

FabianGosebrink:
That is even more than I can ask for. That will be perfect. Thank you!

timdeschryver (approved):
Looks great @FabianGosebrink ! 🏆
It also compiles and works on the project I'm working on.
I do got two questions/comments though:

- I notice that `any` is used here and there, I can pick this up after this PR is merged
- I also noticed that some operations where the config is used uses the following pattern `tokenRefreshInSeconds ?? 0` (I assume this is for if the config option is `null` or `undefined`). Is this correct, because the docs/JSDocs mention a default value. For both cases can something be said, I take the assumption that for these cases we expect the user to configure the option with the `null` value?

projects/angular-auth-oidc-client/src/lib/api/data.service.ts
```diff
@@ -25,15 +25,15 @@ export class DataService {
...
+    url: string | null,
     body: any,
     config: OpenIdConfiguration,
     headersParams?: HttpHeaders
   ): Observable<T> {
     const headers = headersParams || this.prepareHeaders();
     const params = this.prepareParams(config);
 
-    return this.httpClient.post<T>(url, body, { headers, params });
+    return this.httpClient.post<T>(url ?? '', body, { headers, params });
```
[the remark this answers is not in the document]
FabianGosebrink: Yes, I do agree. We should do this in the future.

FabianGosebrink:
> I notice that any is used here and there, I can pick this up after this PR is merged

That would be nice, yes. Happy to help.

> I also noticed that some operations where the config is used uses the following pattern tokenRefreshInSeconds ?? 0 (I assume this is for if the config option is null or undefined). Is this correct, because the docs/JSDocs mention a default value. For both cases can something be said, I take the assumption that for these cases we expect the user to configure the option with the null value?

Ah, we should use the default value of 0. However, what if the user specifies/overwrites it with undefined or null? Well, we could let it crash then, because the value is not supported. Thoughts?

projects/angular-auth-oidc-client/src/lib/callback/interval.service.ts
```diff
@@ -22,7 +22,7 @@ export class IntervalService {
     const millisecondsDelayBetweenTokenCheck = repeatAfterSeconds * 1000;
 
     return new Observable((subscriber) => {
-      let intervalId;
+      let intervalId: NodeJS.Timeout;
```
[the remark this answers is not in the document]
FabianGosebrink: For some reason this does not return a number or id 🤔🤔🤔🤔🤔

timdeschryver:
> Ah, we should use the default value of 0. However, what if the user specifies/overwrites it with undefined or null? Well, we could let it crash then, because the value is not supported. Thoughts?

I believe this can be added in the future with a new major release.

projects/angular-auth-oidc-client/src/lib/callback/interval.service.ts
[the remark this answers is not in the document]
FabianGosebrink: Just fixed it and fixed the tests. Thanks!

FabianGosebrink:
I think we can merge then, can we?

PR #1884, opened 2024-01-03
feat: upgrade to Angular 17

timdeschryver:
I noticed some issues for Angular 17, and I had some time today so I created this PR.
Feel free to close if you were already working on this.
From my understanding this shouldn't be a breaking change.


.github/workflows/build.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 16
+          node-version: 20
```
timdeschryver, line 25: Because Angular dropped support for v16

.eslintrc.json
```diff
@@ -40,7 +40,8 @@
         "@typescript-eslint/ban-types": ["error"],
         "no-useless-escape": ["error"],
         "no-prototype-builtins": ["error"],
-        "prefer-spread": ["error"]
+        "prefer-spread": ["error"],
+        "@typescript-eslint/no-explicit-any": "off"
```
timdeschryver, line 44: It seems like this is enabled by default now, but there were too many issues to address.
Which would make this PR unclear and big, #1848 will resolve this and we can enable this rule back again within that PR.

.github/workflows/build.yml
```diff
@@ -140,15 +140,15 @@ jobs:
         working-directory: ./angular-auth-oidc-client-test
         run: sudo npm run build
 
-  AngularLatestVersionWithStandaloneSchematics:
+  AngularLatestVersionWithNgModuleSchematics:
```
timdeschryver, line 143: The default is updated to `standalone=true`.

.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 {
   "editor.codeActionsOnSave": {
-    "source.organizeImports": true,
-    "source.fixAll.eslint": true
+    "source.organizeImports": "explicit",
```
timdeschryver, line 3: VSCode updated these values automatically. I haven't checked but I assume this is "the new way".

angular.json
```diff
@@ -107,25 +107,25 @@
         "serve": {
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
-            "browserTarget": "sample-code-flow-multi-iframe:build",
             "sslKey": "certs/dev_localhost.key",
             "sslCert": "certs/dev_localhost.pem",
-            "port": 4200
+            "port": 4200,
+            "buildTarget": "sample-code-flow-multi-iframe:build"
```
timdeschryver, line 113: These changes were made by `ng update`.

projects/schematics/src/ng-add/actions/add-standalone-import.ts
```diff
@@ -59,7 +59,7 @@ function addProvider(host: Tree, context: SchematicContext, configName: string,
 
 
   const change = insertImport(
-    source,
+    source as any, // Angular uses the TS 5.1 compiler internally for schematics?
```
timdeschryver, line 62: A quick-fix because I wasn't sure how to fix this the right way.

timdeschryver (commented):
I left some additional comments to highlight some changes.

pavankjadda:
@FabianGosebrink Will there be a new version that includes Angular 17 support?

PR #1919, opened 2024-03-25
refactor: replace any types

timdeschryver:
This PR updates some `any` types to its runtime type or to `unknown` for errors.
Some tests are also updates to replace a "magic string" that contains `any` within the string.


FabianGosebrink:
Nice, thanks!

FabianGosebrink:
Are you finished and we can merge?

timdeschryver:
@FabianGosebrink there are still some `any`'s left, but I already wanted to create this PR (and merge it) because I will be busy the next days.
One side note that I have is that this can result in a breaking change if some of these types are exposed and a user passes an invalid type.

FabianGosebrink:
Let us merge and then see on main and the samples how it goes.

FabianGosebrink:
Thanks!

PR #1936, opened 2024-05-08
Bugfix: Updated URL service `isCallbackFromSts`

### LINKED ISSUE #1935: [Bug]: Query parameter hijacking
### Version

At least 16 and 17

_No response_

```true
The `isCallbackFromSts` only checks for one of the query parameters in the `CALLBACK_PARAMS_TO_CHECK` array, effectively hijacking these parameters for every route in the application.
```

```true
Add any of the `CALLBACK_PARAMS_TO_CHECK` (['code', 'state', 'token', 'id_token']) to the application URL, and the `isCallbackFromSts` method will return true.
```

```true
I would expect the `isCallbackFromSts` to take the redirect URI into consideration.
```

### Additional context

_No response_

erichjsonfosse:
Hey @damienbod :)

Let me know if I've missed or misunderstood anything.
Also let me know if you want a PR for 17 as well.

I added the option to match current URL to the redirect URI when determining whether or not it's a callback.

Fixes #1935 with backwards compatibility.

Also added a comment in the config interface saying `NB: Default will be true in v18.` (for the new `checkRedirectUrlWhenCheckingIfIsCallback` config option), as I do think that this should be done, because the default should be to only regard the redirect URI as a possible callback URL.

### DIFF
--- a/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
+++ b/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
@@ -260,9 +260,10 @@ This is the `redirect_url` which was configured on the Security Token Service (S
 
 - Type: `boolean`
 - Required: `false`
-- Default: `false` *NB:* Default will be `true` in v18.
 
 Whether to check if current URL matches the redirect URI when determining if current URL is in fact the redirect URI.
+
+Default = _false_ *NB:* Default will be `true` in v18.
 
 ### `clientId`
 
--- a/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
+++ b/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
@@ -256,6 +256,14 @@ Allows you to set custom URLs for the Well-Known endpoints.
 
 This is the `redirect_url` which was configured on the Security Token Service (STS).
 
+### `checkRedirectUrlWhenCheckingIfIsCallback`
+
+- Type: `boolean`
+- Required: `false`
+- Default: `false` *NB:* Default will be `true` in v18.
+
+Whether to check if current URL matches the redirect URI when determining if current URL is in fact the redirect URI.
+
 ### `clientId`
 
 - Type: `string`
--- a/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
@@ -73,9 +73,13 @@ export class UrlService {
       const currentUrlInstance = new URL(currentUrl);
       const redirectUriUrlInstance = new URL(this.getRedirectUrl(config));
 
+      const redirectUriWithoutQueryParams = this.getUrlWithoutQueryParameters(redirectUriUrlInstance).toString();
+      const currentUrlWithoutQueryParams = this.getUrlWithoutQueryParameters(currentUrlInstance).toString();
+      const redirectUriQueryParamsArePresentInCurrentUrl = this.queryParametersExist(redirectUriUrlInstance.searchParams, currentUrlInstance.searchParams);
+
       if (
-        this.getUrlWithoutQueryParameters(redirectUriUrlInstance).toString() !== this.getUrlWithoutQueryParameters(currentUrlInstance).toString() ||
-        !this.queryParametersExist(redirectUriUrlInstance.searchParams, currentUrlInstance.searchParams)
+        redirectUriWithoutQueryParams !== currentUrlWithoutQueryParams ||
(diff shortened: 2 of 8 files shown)


projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -40,7 +40,47 @@ export class UrlService {
...
+      u.searchParams.delete(key);
+    });
+
+    return u;
+  }
+
+  queryParametersExist(expected: URLSearchParams, actual: URLSearchParams): boolean {
+    let r = true;
+
+    expected.forEach((v, k) => {
```
FabianGosebrink, line 62: Can we work with a `some` instead of a `foreach` and a variable? Have not tried it out.

projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -40,7 +40,47 @@ export class UrlService {
...
+    return r;
+  }
+
+  isCallbackFromSts(currentUrl: string, config?: OpenIdConfiguration): boolean {
+    if (config && config.checkRedirectUrlWhenCheckingIfIsCallback) {
+      const currentUrlInstance = new URL(currentUrl);
+      const redirectUriUrlInstance = new URL(this.getRedirectUrl(config));
+
+      if (
+        this.getUrlWithoutQueryParameters(redirectUriUrlInstance).toString() !== this.getUrlWithoutQueryParameters(currentUrlInstance).toString() ||
```
FabianGosebrink, line 77: simplify this if-statement with variables to make it bette readable.

FabianGosebrink (changes requested):

FabianGosebrink:
Thanks for this PR, that is looking really nice. Thank you! Can we also update the docs accordingly?

projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
erichjsonfosse: The `URLSearchParams` interface does not seem to have a `some`-method

erichjsonfosse:
@damienbod 
I've added a section for `checkRedirectUrlWhenCheckingIfIsCallback` to the `configuration.md`-file. Was that the kind of documentation you were looking for?

erichjsonfosse:
@FabianGosebrink 
I've added a section for `checkRedirectUrlWhenCheckingIfIsCallback` to the `configuration.md`-file. Was that the kind of documentation you were looking for?

docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
```diff
@@ -256,6 +256,14 @@ Allows you to set custom URLs for the Well-Known endpoints.
 
 This is the `redirect_url` which was configured on the Security Token Service (STS).
 
+### `checkRedirectUrlWhenCheckingIfIsCallback`
+
+- Type: `boolean`
+- Required: `false`
+- Default: `false` *NB:* Default will be `true` in v18.
```
timdeschryver, line 263: To keep things consistent with the rest of the docs, the default should be moved to below the description. For example see `silentRenewTimeoutInSeconds`

erichjsonfosse: @timdeschryver 
Good point!
Better now? :)

FabianGosebrink (approved):

docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
timdeschryver: Thanks @erichjsonfosse

timdeschryver (approved):

erichjsonfosse:
Thanks for helping out here, everyone!

Any ETA on an updated package :) ?

erichjsonfosse:
> Thanks for helping out here, everyone!
> 
> Any ETA on an updated package :) ?

@FabianGosebrink @timdeschryver @damienbod :)

erichjsonfosse:
@FabianGosebrink @timdeschryver @damienbod 

Is there anything I can do to help speed up the deployment of an updated 16 version?

damienbod:
Hi @erichjsonfosse releasing a  new package for v16 was not on my radar. Let me check.

damienbod:
16.0.2 released

erichjsonfosse:
@damienbod 
Do you need me to fire up a PR for the 17 and 18 versions as well, or do you have what you need in this one :) ?

damienbod:
Version 18 would be nice, thanks

PR #1937, opened 2024-05-10
feat: add authenticated and userData signals

timdeschryver:
I had some time so I wanted to try-out signals within this library.

This PR adds two signals for the most common/used methods (from my perspective), which are `isAuthenticated$` and `userData$`. I've just wrapped the observable with `toSignal`.

I've updated the standalone example app to test it out.

Thoughts? I can also convert this into an issue.

Anonymous:
image

Authenticated:
image


FabianGosebrink:
Thanks for this PR. I think it is a nice addition to move otwards signals. Thanks!!!

PR #1943, opened 2024-05-17
fix(refresh-session): `forceRefreshSession` does not reset `storageSilentRenewRunning`

### LINKED ISSUE #1942: [Bug]: `forceRefreshSession` does not reset `storageSilentRenewRunning`
### Version

17.0.0

_No response_

When using `forceRefreshSession` it set `storageSilentRenewRunning` in the storage.

The problem is that when the refresh finish, the value is not reset, meaning that we can't re-use the `forceRefreshSession`.

Currently, when using `forceRefreshSession`, the only way that `storageSilentRenewRunning` is reset is to wait for the `periodicallyCheck` (https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/callback/periodically-token-check.service.ts#L105).

This also mean that if we set `tokenRefreshInSeconds: 60`, we will have to wait a lot of time before being able to re-use `forceRefreshSession`.

I'm using this function to refresh the session when I update the user's profile. The `user` object in my app is constructed with the `userData` that I get from event `UserDataChanged`.

- https://github.com/damienbod/angular-auth-oidc-client/tree/main/projects/sample-code-flow-refresh-tokens
- set `tokenRefreshInSeconds: 60`
- trigger `forceRefreshSession`
- There will be a value for `storageSilentRenewRunning` in the storage
- Need to wait for the `periodicallyCheck` until this is reset

Once `forceRefreshSession` is called, it should call `resetSilentRenewRunning`.

### Additional context

I will provide a PR.

Yberion:
Hello,

This PR fixes #1942.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts
@@ -8,8 +8,6 @@ import {
   timer,
 } from 'rxjs';
 import {
-  catchError,
-  finalize,
   map,
   mergeMap,
   retryWhen,
@@ -27,8 +25,6 @@ import { RefreshSessionIframeService } from '../iframe/refresh-session-iframe.se
 import { SilentRenewService } from '../iframe/silent-renew.service';
 import { LoggerService } from '../logging/logger.service';
 import { LoginResponse } from '../login/login-response';
-import { EventTypes } from '../public-events/event-types';
-import { PublicEventsService } from '../public-events/public-events.service';
 import { StoragePersistenceService } from '../storage/storage-persistence.service';
 import { UserService } from '../user-data/user.service';
 import { FlowHelper } from '../utils/flowHelper/flow-helper.service';
@@ -43,8 +39,6 @@ export class RefreshSessionService {
   private readonly flowsDataService = inject(FlowsDataService);
 
   private readonly loggerService = inject(LoggerService);
-
-  private readonly publicEventsService = inject(PublicEventsService);
 
   private readonly silentRenewService = inject(SilentRenewService);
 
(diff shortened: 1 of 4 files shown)


projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts
```diff
@@ -75,7 +82,21 @@ export class RefreshSessionService {
 
     this.persistCustomParams(extraCustomParams, config);
 
-    return this.forceRefreshSession(config, allConfigs, extraCustomParams);
+    // Best place ?
+    this.publicEventsService.fireEvent(EventTypes.ManualRenewStarted);
```
Yberion, line 86: Is it the best place?

projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts
```diff
@@ -143,6 +164,7 @@ export class RefreshSessionService {
               `forceRefreshSession timeout. Attempt #${currentAttempt}`
             );
 
+            // Still needed ?
             this.flowsDataService.resetSilentRenewRunning(config);
```
Yberion, line 168: Is it still needed there as we call it in the finalize line 95?

projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts
```diff
@@ -143,6 +164,7 @@ export class RefreshSessionService {
               `forceRefreshSession timeout. Attempt #${currentAttempt}`
             );
 
+            // Still needed ?
```
timdeschryver, line 167: What's the reasoning behind the new events?
I think it can also be solved by invoking `resetSilentRenewRunning` within the callback at https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts#L94. Similar to the current approach at https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts#L146

Yberion: Now that you ask me and after thinking a bit more, as `forceRefreshSession` returns an Observable, this is indeed useless to create those new events, as we can achieve the same right now.

I will remove the new events.

On why I put `resetSilentRenewRunning` line 95 in a `finalize` (`tap` when I'll remove the events), it's because I prefer to not duplicate logic, that's why I was wondering if the one line 167 was still needed, but I think it's not anymore.

But both are fine for me, as it will fix the issue anyway.

Any "preference"?

Yberion: Looking a bit more, I think, if we hit this line https://github.com/damienbod/angular-auth-oidc-client/blob/main/projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.ts#L138, we will face the same problem, not being able to re-use `forceRefreshSession` as `resetSilentRenewRunning ` will not be called

timdeschryver (approved):
From my point of view this change looks good 👍
Side note: if you force the refresh token very rapidly (when a request is already pending) then this results in an error "0-angularCodeRefreshTokens - no refresh token found, please login", but I think this behavior was already there and isn't caused by this change.

Issue #1954, opened 2024-06-04
[Bug]: Missing `EventTypes.CheckingAuthFinished` for state isAuthenticated: false


FestimShyti:
### Version

7.1.0

### Please provide the exception or error you saw

```true
File: `check-auth.service.ts`

if (isAuthenticated) {
    this.startCheckSessionAndValidation(config, allConfigs);

    if (!isCallback) {
        this.authStateService.setAuthenticatedAndFireEvent(allConfigs);
        this.userService.publishUserDataIfExists(config, allConfigs);
    }

    // This event moved here and fired only isAuthenticated: true
    this.publicEventsService.fireEvent(EventTypes.CheckingAuthFinished);
}
.
.
.

tap(({ isAuthenticated }) => {
    // Fire event should be here to handle both states for isAuthenticated...

    if (isAuthenticated) {
        this.autoLoginService.checkSavedRedirectRouteAndNavigate(config);
    }
}),
```

### Steps to reproduce the behavior

```true
When authenticated is false the flow does not trigger `EventTypes.CheckingAuthFinished`.
```

PR #1961, opened 2024-06-10
Fix issue #1954: Ensure CheckingAuthFinished event fires regardless of authentication state

jagipas:
Hello,

This pull request addresses issue [#1954](https://github.com/damienbod/angular-auth-oidc-client/issues/1954) by ensuring that the `EventTypes.CheckingAuthFinished` event is fired regardless of the authentication state.

### Problem

In commit [4143e2444330dc86ab3a59574f7511b461e75a61](https://github.com/damienbod/angular-auth-oidc-client/commit/4143e2444330dc86ab3a59574f7511b461e75a61), the `fireEvent(EventTypes.CheckingAuthFinished)` was moved inside the `if (isAuthenticated)` block. This change caused the event not to fire when the authentication state is false, which leads to the issue described in #1954.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/auth-state/check-auth.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/auth-state/check-auth.service.spec.ts
@@ -608,6 +608,28 @@ describe('CheckAuthService', () => {
         ]);
       });
     }));
+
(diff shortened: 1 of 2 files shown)


joewIST:
Any idea when this will be released? We will need to revert back down to v17 for now

damienbod:
I will release tomorrow, will do some testing

PR #1976, opened 2024-07-05
Bugfix: Updated URL service `isCallbackFromSts`

erichjsonfosse:
Added the option to match current URL to the redirect URI when determining whether or not it's a callback.

Fixes https://github.com/damienbod/angular-auth-oidc-client/issues/1935

As per the [comment](https://github.com/damienbod/angular-auth-oidc-client/blob/c0b77febe774334dff17a878b1e5376bec3a14f3/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md?plain=1#L266) in the v16 changes, the default value for `checkRedirectUrlWhenCheckingIfIsCallback` is now `true`.

A pre-push hook for Prettier ran when I pushed my changes, so around 30 of the changed files are a result of that.


erichjsonfosse:
@damienbod @FabianGosebrink 

Anything else needed on this one? :)

FabianGosebrink:
Hey, thanks for this PR. I did not see this, yet. Will have a look tomorrow. Thanks!!!

projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -44,7 +44,74 @@ export class UrlService {
...
+    }
+
+    keys.forEach((key) => {
+      u.searchParams.delete(key);
+    });
+
+    return u;
+  }
+
+  queryParametersExist(
```
FabianGosebrink, line 63:
```suggestion
  queryParametersExist(expected: URLSearchParams, actual: URLSearchParams): boolean {
  const expectedKeys = Array.from(expected.keys());
  
  return  expectedKeys.every(key => actual.has(key));
}
```

FabianGosebrink (changes requested):

FabianGosebrink:
Small suggestion for beautifying the code, otherwise this is totally fine. Thank you very much!

erichjsonfosse:
@FabianGosebrink looks like the "commit suggestion" feature didn't do well here, so I'm fixing locally now. Just a sec :)

erichjsonfosse:
@FabianGosebrink @damienbod 
Do you know when a new version with this fix might be available?

PR #2000, opened 2024-09-17
feat: add guard `autoLoginPartialRoutesGuardWithConfig` for specific configurations

kyubisation:
It is currently not possible to use the login guard for multiple configurations, as the current guard will only check the first configuration. This PR adds `autoLoginPartialRoutesGuardWithConfig`, which accepts a parameter and returns a guard function.

```ts
  ...
  {
    ...
    canActivate: [autoLoginPartialRoutesGuardWithConfig('myConfig')],
  },
```


FabianGosebrink:
Thanks for the PR and the update of the GH Actions :)

PR #2006, opened 2024-09-18
feat: add `withAppInitializerAuthCheck` as a feature for `provideAuth`

### LINKED ISSUE #1926: [Feature Request]: Option for authCheck during app initialization phase
**Is your feature request related to a problem? Please describe.**
Currently the `OidcSecurityService.checkAuth()` is recommended to be called in the AppComponent, which is a sensible default.
However I would prefer to have an option to call it during app initialization (with a provider for APP_INITIALIZER).

**Describe the solution you'd like**
I would like to have an optional possibility to run the auth check during the app initialization phase.

e.g.
```ts
export function withAppInitializerAuthCheck() {
  return {
    provide: APP_INITIALIZER,
    useFactory: (oidcSecurityService: OidcSecurityService) => () =>
      oidcSecurityService.checkAuthMultiple(),
    multi: true,
    deps: [OidcSecurityService],
  };
}

...
provideAuth(authConfig, withAppInitializerAuthCheck()),
...

// or alternatively

export interface PassedInitialConfig {
    config?: OpenIdConfiguration | OpenIdConfiguration[];
    loader?: Provider;
    appInitializerAuthCheck?: boolean;
}
```

**Describe alternatives you've considered**
Just use the code above in my apps.

**Additional context**
I am aware that this is mostly a convenience functionality.
Due to this I am more that willing to contribute the necessary code, if this feature is an acceptable addition and there is agreement on the API.

kyubisation:
This PR adds a feature `withAppInitializerAuthCheck` to handle OAuth callbacks during app initialization phase. This replaces the need to manually call `OidcSecurityService.checkAuth(...)` or `OidcSecurityService.checkAuthMultiple(...)`.

Closes #1926


FabianGosebrink:
Thanks, that is a nice idea. Love it. This would also mean that we have to update the samples, is this correct?

kyubisation:
Thank you 😃 
As it is an opt-in, I was not sure how deep to integrate it into existing code. Is there a specific example or range of examples where you would like to see this used?

FabianGosebrink:
We have to thank YOU for this PR ❤️ 

It would be nice to see the either/or comparison in the docs, because the users have to erase the `checkAuth(...)` step, which gets forgotten really easily. However, maybe we can enhance an existing sample "sample-code-flow-azuread" or "sample-code-flow-auth0", comment out the added code and comment it like you did on the docs, so that the users can see what they would have to change. What do you think?

kyubisation:
I think that's fine. 👍 
I'm also fine with creating a new example, if that is preferred.
I'll have a look to adapt this, but it might be next week.

Maybe one question is; Should this potentially become the recommendation (in examples) in the future? While we at SBB have been using this approach for a while, I am hesitant to suggest a change in the general approach, as I am not a maintainer of this project. We can also wait a while before evaluating whether a change would be an improvement.

FabianGosebrink:
Generally: Yes, I would recommend this. In practice and out of experience, I know that auth and so also this lib is used in ways you can not even imagine. So if it fits for EVERY case? I doubt it, because auth is very special. However, we want to provide a go to and fast solution to do your auth, so your feature is self explaining and with the docs we can also advise for the alternative solution. If we see that your feature establishes as stable, when can absolutely take it as the recommended solution. I, personally, will absolutely use it that way :)

kyubisation:
That sounds reasonable. 👍 
And yes, I am unfortunately very much aware of the many ways auth can be used (and misused).

No, not at the moment, but that statement might be misleading.
We previously used the keycloak library or https://www.npmjs.com/package/angular-oauth2-oidc. This year we evaluated this library (and others), with the result that this library matches our use case the best.
Due to this, it is now the recommended authentication library for new projects. but that does not affect our hundreds of existing Angular applications that might or might not migrate the auth library.
I hope that puts it into perspective.

FabianGosebrink:
Hey, thanks. Yeah, that is what I also see at customer projects. New project use the new guidelines, and old ones might or might not be migrated. Nice. Let me know if you are missing any feature, have a question or need something. We leave this PR for now and

- [ ] enhance one sample with the new way commented out
- [ ] add docs to have the "old" and "new" way side by side or underneath, so that we can link to it

Your PR is awesome. Thanks so much.

kyubisation:
@FabianGosebrink I just realized that currently the only example using `provideAuth` is the `sample-code-flow-standalone`. Should I enhance that example or should e.g. the azuread example be migrated to standalone?

docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
```diff
@@ -219,6 +219,24 @@ export const appConfig: ApplicationConfig = {
...
+
+...
+    provideAuth({
+      config: {
+        /* Your config here */
+      },
+    },
+    withAppInitializerAuthCheck()),
+...
+```
```
timdeschryver, line 238: I think it will be more clear if we provide a full example here.

```suggestion
```ts
import { ApplicationConfig } from '@angular/core';
import { provideAuth, withAppInitializerAuthCheck } from 'angular-auth-oidc-client';

export const appConfig: ApplicationConfig = {
  providers: [
    provideAuth({
      config: {
        /* Your config here */
      },
      withAppInitializerAuthCheck()
    }),
  ],
};

```
```

**Review by timdeschryver — COMMENTED** [review]
I like this feature!

I think updating the standalone example is fine for now, as we can use it to verify that this feature works as expected.
 
With standalone becoming the default I think we should update all examples in seperate PR(s).
We can leave one using `NgModule` to also cover.
I suppose the docs will need some enhancements too.

File: docs/site/angular-auth-oidc-client/docs/documentation/configuration.md (hunk 1 above)
Comment (kyubisation) [reply docs/site/angular-auth-oidc-client/docs/documentation/configuration.md:238]: Will do 👍

**Comment by kyubisation** [comment]
I have migrated the azuread example, as this leaves an example with the current/previous standalone configuration. I have also extended the documentation a little bit, as discussed.
Feel free to have another look and tell me, if I still need to address something.

File: docs/site/angular-auth-oidc-client/docs/documentation/configuration.md [hunk 2]
```diff
@@ -219,6 +219,51 @@ export const appConfig: ApplicationConfig = {
...
+  ngOnInit(): void {
+    this.oidcSecurityService
+      .checkAuth()
+      .subscribe(({ isAuthenticated, accessToken }) => {
+        console.log('app authenticated', isAuthenticated);
+        console.log(`Current access token is '${accessToken}'`);
+      });
+  }
+...
+```
```
timdeschryver, line 265: We can also refer to the `checkAuth` API, this keeps the documentation in one place.

timdeschryver (approved):
LGTM 👍
The comment is a personal one, but isn't "blocking".

docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
kyubisation: Could you explain this in more detail? I'm not sure I fully understand what you mean.

kyubisation: I can do that, but @FabianGosebrink requested to `add docs to have the "old" and "new" way side by side or underneath, so that we can link to it`.
I'm fine with either. Should I adapt, as suggested by @timdeschryver?

timdeschryver: Oh nevermind, my comment then :)

FabianGosebrink: Easy, I just wanted it to make it as easy as possible for the users to use. I am fine with both :)

FabianGosebrink:
Thank you both for being so active here :)

kyubisation:
Yes, I was merely waiting for any advice on how to proceed with the raised issue.
If it is fine, as it is, it can be merged. If I should adapt the documentation, I can do that if desired.

FabianGosebrink:
Thanks :)

kyubisation:
Thank you both for the review, feedback and merge. 😃 👍

FabianGosebrink:
Thank YOU for the work you did!! This lib grows because of contributions from people like you. You did amazing. Thanks a million!

konstantinvlasenko:
Question. I am using 18.0.1
I don't see **withAppInitializerAuthCheck** 
Am I missing something?

kyubisation:
This feature has not been released yet. I am not overly familiar with the release cycles of this library, but I assume this will be released with version 19 in November.

FabianGosebrink:
Yes, we will run the last tests and then release with this feature included. Thanks!

damienbod:
I will do some testing, docs and try to release this weekend

Greetings and thanks for the PR, Damien

goncard:
Question: when we are guarding all the routes with AutoLoginPartialRoutesGuard, will this feature replace the need to call checkAuth on app component or callback component? I am calling it on the callback component which is my public route but if it will not be needed anymore, the callback component will basically be useless.

PR #2014, opened 2024-09-27
fix: prevent guard to be called with wrong parameter

kyubisation:
Previously `autoLoginPartialRoutesGuardWithConfig` was added/implemented, which added a parameter to `autoLoginPartialRoutesGuard`. This however introduced a subtle bug, as guards are called with a second parameter, which broke the usage.
This change extracts the common code into its own function and removes the second parameter from `autoLoginPartialRoutesGuard`.


kyubisation:
Sorry, just realized that this was already solved and I was working on an outdated state.

PR #2029, opened 2024-10-18
docs: update auto-login page

timdeschryver:


docs/site/angular-auth-oidc-client/docs/documentation/auto-login.md
```diff
@@ -5,102 +5,102 @@ sidebar_position: 9
...
-    this.oidcSecurityService
-      .checkAuth()
-      .subscribe(({ isAuthenticated, userData, accessToken }) => {
-        // ...
-      });
-  }
-}
-```
-
-### Auto Login when all routes are guarded
```
timdeschryver, line 62: I removed this because it's deprecated.
If wanted I can re-add it with the NgModule example.

docs/site/angular-auth-oidc-client/docs/documentation/auto-login.md
```diff
@@ -5,102 +5,102 @@ sidebar_position: 9
...
   },
   { path: 'unauthorized', component: UnauthorizedComponent },
 ];
 ```
 
 In this case, the `/home` and `/unauthorized` routes are not protected, and they are accessible without a login.
-Please make sure to call `checkAuth()` like normal in your `app.component.ts`.
-
-```ts
-export class AppComponent implements OnInit {
```
timdeschryver, line 49: I removed this because we also have the `withAppInitializerAuthCheck`, and I didn't want to bloath these docs with adding an example with `withAppInitializerAuthCheck`.

timdeschryver (commented):
I've updated the page to:
- use functional guards by default
- update angular links to angular.dev
- updated the examples to use routes instead of NgModule

PR #2046, opened 2024-11-27
Silent renew prompt mode is now configurable via options.

### LINKED ISSUE #1856: [Question]:
I am using version 14.1.5

**Question**
How can I use implicit flow with a login prompt and an iframe silent refresh?
I have configured the library like this `customParamsAuthRequest: { prompt : 'login' }` because I am authenticating my users against the ADFS and I need to force the web form instead of NTLM.
When I use `oidcService.checkAuth()`, I have no silent refresh, and when I use `oidcService.checkAuthIncludingServer()` my `prompt : 'login'` parameter gets arbitrarily overridden with a `prompt : 'none'` in the `createAuthorizeUrl` function.

jonathanantoine:
Allow to configure the prompt option used when performing a silent renew. 

Helps to fix #1856

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
@@ -408,7 +408,7 @@ export class UrlService {
     const authWellKnownEndPoints = this.storagePersistenceService.read('authWellKnownEndPoints', configuration);
 
     if (authWellKnownEndPoints) {
-      return this.createAuthorizeUrl('', silentRenewUrl, nonce, state, configuration, 'none', customParams);
+      return this.createAuthorizeUrl('', silentRenewUrl, nonce, state, configuration, configuration.silentRenewPrompt, customParams);
     }
 
     this.loggerService.logError(configuration, 'authWellKnownEndpoints is undefined');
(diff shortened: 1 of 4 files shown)


projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
```diff
@@ -439,7 +439,7 @@ export class UrlService {
         const authWellKnownEndPoints = this.storagePersistenceService.read('authWellKnownEndPoints', configuration);
 
         if (authWellKnownEndPoints) {
-          return this.createAuthorizeUrl(codeChallenge, silentRenewUrl, nonce, state, configuration, 'none', customParams);
+          return this.createAuthorizeUrl(codeChallenge, silentRenewUrl, nonce, state, configuration,configuration.silentRenewPrompt, customParams);
```
FabianGosebrink, line 442: We are passing the configuration into the method already. No need for this parameter and we can extract the param in the method, can't we?

FabianGosebrink (changes requested):

projects/angular-auth-oidc-client/src/lib/utils/url/url.service.ts
jonathanantoine: I want to be sure its only applied when in silent renew mode so I think its better this way.

FabianGosebrink: Thanks for your comment. I do not understand why this would be the case when reading the parameter in the method. Can you elaborate?

FabianGosebrink (approved):

FabianGosebrink:
Why is the target branch V14?

jonathanantoine:
I need it in v14 version but I plan to create the other merging PRs once this one is approved.

Thanks for the approval @FabianGosebrink 

(What is missing for this to be merged please ?)😄

jonathanantoine:
Thank you 💖

Issue #2040, opened 2024-11-01
[Bug]: isCallbackFromSts doesn't respect trailing slash in the url


rammba:
### Version

18.0.2

### Please provide the exception or error you saw

```true
No exceptions nor errors, just failed login flow.
```

### Steps to reproduce the behavior

```true
My app is using Azure B2C authentication provider. In the B2C configuration I have auth redirectUrl without trailing slash, but looks like B2C adds it implicitly.
```

PR #2062, opened 2025-01-09
add debug log for redirectUrl check

gobeli:
Just ran into this, can be tricky to find out what is wrong if it just aborts without a log.


FabianGosebrink:
Thank you!

rammba:
This change logs error which happened in #2040

Issue #2015, opened 2024-09-28
[Bug]: Multiple Event Listeners Accumulation in CheckSessionService


jagipas:
### Version

18.0.1

### Please provide the exception or error you saw

```true
## Description
The `CheckSessionService` is accumulating multiple event listeners for the 'message' event, causing the `messageHandler` to be executed multiple times for each event. This issue appears to have been introduced in PR #1735.

- Each time `bindMessageEventToIframe()` is called, a new event listener is added to `defaultView` without removing the previous one.
- This results in multiple executions of the `messageHandler` for a single 'message' event.
- The problem worsens over time as more listeners accumulate.
```

### Steps to reproduce the behavior

```true
1. Initialize the `CheckSessionService`.
2. Call `bindMessageEventToIframe()` multiple times (this may happen implicitly through other service methods).
3. Observe that 'message' events trigger multiple executions of the `messageHandler`.
```

### A clear and concise description of what you expected to happen.

```true
- Only one event listener should be active at any given time.
- When `bindMessageEventToIframe()` is called, it should remove any existing listener before adding a new one.
```

## Possible Solution
Modify the `bindMessageEventToIframe()` method to remove the existing listener before adding a new one:

```typescript
private bindMessageEventToIframe(configuration: OpenIdConfiguration): void {
  const defaultView = this.document.defaultView;

  // Remove existing listener if there is one
  if (this.iframeMessageEventListener && defaultView) {
    defaultView.removeEventListener('message', this.iframeMessageEventListener, false);
  }

  // Create and assign the new listener
  this.iframeMessageEventListener = this.messageHandler.bind(
    this,
    configuration
  );

  if (defaultView) {
    defaultView.addEventListener(
      'message',
      this.iframeMessageEventListener,
      false
    );
  }
}
```

PR #2074, opened 2025-02-06
Prevent event listener accumulation

florianlevy:
Fix event listener leak in CheckSession service

After migrating our applications to Angular 17, users reported significant performance degradation after 2-3 hours of application usage. Our analysis revealed abnormal CPU usage and an increasing number of event listeners without any user actions.

This issue was previously reported in [#2015.](https://github.com/damienbod/angular-auth-oidc-client/issues/2015) Implementing the proposed solution on fork resolved the performance problems.

Changes made:
- Properly remove existing event listeners before adding new ones in CheckSession service
- Prevent event listener accumulation during session checks
- Add unit test

This fix addresses the eventListener leak that was causing performance degradation over time.

Related issue: [#2015](https://github.com/damienbod/angular-auth-oidc-client/issues/2015)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
         run: ls
 
       - name: Upload Artefact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: angular_auth_oidc_client_artefact
           path: dist/angular-auth-oidc-client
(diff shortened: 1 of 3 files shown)


FabianGosebrink:
Omg thanks so much!

florianlevy:
Hi, I pushed another commit to resolve the build error due to the deprecation of the upload and download actions. I updated all actions to v4.

PR #2080, opened 2025-03-06
feat(interceptor): allow wildcard securedRoutes

### LINKED ISSUE #2079: [Question]: Is it possible to have wildcards in secureRoutes setup
19.0.0

**Question**
Is it possible to define a wildcard for the secured routes? Something like below

```
{
  ...
  secureRoutes: ['https://my-secure-url.com/*/token']
}
```

skatterwe:
enables wildcards for secureRoute configuration.

Fixes #2079


skatterwe:
Docs are still missing. Will have a look at that later. And maybe you have a better idea regex wise so that we can combine startWith into the regex. But I wasn't able to find one yet that matched all test cases

skatterwe:
@FabianGosebrink  added the documentation. Can you please check if you think this is sufficient. Also I'm interested on your thoughts about the regex check. Do you have an idea if there is a nice regex that would fit both cases, so we don't need the startWith comparison? I'm not a huge regex expert and didn't manage to find anything suitable yet.

FabianGosebrink:
I refactored a bit and tried to improve the regex. Can you have a look, give me your "OK" and then we can emrgen :-)

FabianGosebrink:
@damienbod : Waiting for @skatterwe :-)

skatterwe:
@FabianGosebrink  sorry was away for a while. looks fine to me

FabianGosebrink:
Can you resolve the conflicts? Should only be ESLint stuff.

skatterwe:
@FabianGosebrink  done

FabianGosebrink:
Awesome, thanks!

damienbod:
I'll make a minor release for this

PR #2091, opened 2025-04-10
docs: fix import for standalone quickstart

miguelerm:
just a quick fix for the standalone docs


timdeschryver (approved):
Good catch! 👍

FabianGosebrink:
Thanks!

PR #2098, opened 2025-05-12
feat(core): upgrade to Angular 20 and refactor imports

pavankjadda:


projects/angular-auth-oidc-client/src/lib/callback/interval.service.ts
```diff
@@ -1,6 +1,6 @@
-import { inject, Injectable, NgZone } from '@angular/core';
+import { inject, Injectable, NgZone, DOCUMENT } from '@angular/core';
```
timdeschryver, line 1: These changes have caused the build to fail for earlier versions of Angular.
Is this change required or can we keep it as is?

If we can't keep it, this is a breaking change and:
- needs to be documented
- the build pipeline needs to be updated

projects/angular-auth-oidc-client/src/lib/flows/callback-handling/history-jwt-keys-callback-handler.service.ts
```diff
@@ -1,5 +1,5 @@
-import { DOCUMENT } from '@angular/common';
-import { inject, Injectable } from '@angular/core';
+
```
timdeschryver, line 1: Some files start with a new line, which needs to be reverted:
```suggestion
```

PR #2116, opened 2025-08-12
feat: strict issuer validation on OIDC Discovery document retrieval

syalioune:
## Feature: Implement Issuer Validation for OpenID Connect Discovery document

This PR proposes to enhance the security of `angular-auth-oidc-client` by implementing a validation check for the `issuer` field in the OpenID Connect discovery document.

**Problem:**

During a recent internal review, we identified a scenario where a misconfigured reverse proxy could allow a cache poisoning attack. This attack could lead to a modified `.well-known/openid-configuration` payload being served from a malicious domain (e.g., `evil.xyz.com`), even when the configured authority is a legitimate Identity Provider (e.g., `idp.corporate.com`).

Currently, `angular-auth-oidc-client` accepts this malicious metadata without validating that the `issuer` field within the discovery document matches the expected authority.

**Proposed Solution:**

According to the OpenID Connect Discovery specification (Section 4.3):

> "The issuer value returned MUST be identical to the Issuer URL that was used as the prefix to /.well-known/openid-configuration to retrieve the configuration information.”

This PR aims to implement this crucial validation. By verifying that the `issuer` value returned in the discovery document precisely matches the authority configured in the client, we can provide a significant defense-in-depth layer against cache poisoning and similar attacks.

**Benefits:**

*   **Enhanced Security:** Prevents the library from accepting forged or malicious OpenID Connect configurations.
*   **Specification Compliance:** Ensures the library adheres strictly to the OpenID Connect Discovery specification.
*   **Defense-in-Depth:** Adds an important security control, even if upstream infrastructure issues are present.

**Workaround :**

Using the `authWellknownEndpoints` configuration option to explicitly pin the well-known endpoint URLs. However, we believe this check would benefit all users of the library as a standard security measure.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/config/auth-well-known/auth-well-known-data.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/config/auth-well-known/auth-well-known-data.service.spec.ts
@@ -24,6 +24,7 @@ const DUMMY_WELL_KNOWN_DOCUMENT = {
   introspection_endpoint:
     'https://identity-server.test/realms/main/protocol/openid-connect/token/introspect',
 };
+const DUMMY_MALICIOUS_URL = 'https://malicious.test/realms/main';
 
 describe('AuthWellKnownDataService', () => {
   let service: AuthWellKnownDataService;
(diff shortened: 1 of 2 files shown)


damienbod:
I had to revert this. It breaks the **start-sample-code-flow-multi-Auth0-ID4** sample

multiple IDPs

PR #2141, opened 2025-10-02
Fixed wrong DOCUMENT import. Change from @angular/core to @angular/common

najack60:
I was getting build errors that were happening because DOCUMENT was getting imported from @angular/core instead of @angular/common. This PR addresses that issue.


najack60:
> What version of Angular are you on (you will need v20.x)?

I'm on version: 19.1.7. Are you saying this could be fixed by upgrading to v20.x, or that I need to be on version 20.x to contribute?

PR #2166, opened 2025-12-31
fix(public) : Export PopupResult to public

PikPakPik:
export public PopupResult


FabianGosebrink:
Thanks for the PR. Are you referencing to an issue here?

PikPakPik:
Yes, I have a problem with it. Following a migration of the TypeScript, nx, and Angular packages, I had a problem: the project wouldn't launch because it couldn't find the type.

PR #2186, opened 2026-03-12
Fix stale tokens returned during concurrent refresh-token renewal

ThorstenKunz:
## Summary

This PR fixes a race condition in the refresh-token flow where concurrent `forceRefreshSession()` calls could return stale tokens or observe inconsistent auth state.

## Problem

In refresh-token mode, `forceRefreshSession()` built its `LoginResponse` from the current auth-state getters instead of preferring the fresh result from the completed refresh operation.

This became problematic when multiple refresh calls happened close together:

- one call could trigger the actual refresh,
- another call could see that renew was already running,
- and callers could still receive stale tokens from storage instead of the newly refreshed values.

## Reproduction

The issue can be reproduced with a client configured with `useRefreshToken: true` by triggering multiple manual refresh calls in parallel:

```ts
await Promise.all([
  firstValueFrom(oidcSecurityService.forceRefreshSession(undefined, configId)),
  firstValueFrom(oidcSecurityService.forceRefreshSession(undefined, configId)),
]);
```

Before the fix, returned tokens and immediately readable stored tokens could diverge depending on timing.

## Fix

The refresh flow was changed in three key places:

1. The refresh-token path now prefers the `authResult` of the completed refresh operation when creating the `LoginResponse`.
2. If a renew process is already running, additional callers now wait for the published `NewAuthenticationResult` event before completing.
3. The renew-running flag is set earlier, before the asynchronous well-known endpoint lookup, to close a timing window for concurrent refresh requests.

## Test Coverage

This PR adds regression coverage for both unit and integration scenarios:

- unit tests for stale token reads and concurrent refresh handling,
- an integration test covering concurrent `forceRefreshSession()` calls with `useRefreshToken: true`,
- test IDP updates to support refresh tokens and refresh-token rotation.

## Result

After this change, concurrent manual refresh calls resolve consistently and return the freshly renewed token set instead of stale auth-state values.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/callback/refresh-session.service.spec.ts
@@ -1,5 +1,5 @@
 import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
-import { of, throwError } from 'rxjs';
+import { of, Subject, throwError } from 'rxjs';
 import { delay } from 'rxjs/operators';
 import { mockProvider } from '../../test/auto-mock';
 import { AuthStateService } from '../auth-state/auth-state.service';
@@ -10,6 +10,7 @@ import { RefreshSessionIframeService } from '../iframe/refresh-session-iframe.se
 import { SilentRenewService } from '../iframe/silent-renew.service';
 import { LoggerService } from '../logging/logger.service';
 import { LoginResponse } from '../login/login-response';
+import { EventTypes } from '../public-events/event-types';
 import { PublicEventsService } from '../public-events/public-events.service';
 import { StoragePersistenceService } from '../storage/storage-persistence.service';
 import { UserService } from '../user-data/user.service';
@@ -30,6 +31,8 @@ describe('RefreshSessionService ', () => {
   let refreshSessionIframeService: RefreshSessionIframeService;
   let refreshSessionRefreshTokenService: RefreshSessionRefreshTokenService;
   let authWellKnownService: AuthWellKnownService;
+  let publicEventsService: PublicEventsService;
+  let userService: UserService;
 
   beforeEach(() => {
     TestBed.configureTestingModule({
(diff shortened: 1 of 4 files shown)


ThorstenKunz:
Since this PR has not received further review yet, I asked an OpenAI 5.6 Sol coding agent to analyze the reported race condition and look for additional improvements that might help move the PR forward. The analysis identified another potential issue, and I added a follow-up commit addressing it.
Here is a summary of the production incident and the additional changes:
I investigated a recent production incident that provides further evidence for the race condition addressed by this PR.
In that incident, a periodic silent renew raced with a manual forceRefreshSession() call after the browser resumed with an expired access token. Both refresh paths passed the silentRenewRunning check before the flag was set, resulting in two discovery/JWKS requests and two token requests.
Both token requests returned HTTP 200 with fresh access and refresh tokens. However, the concurrent callback processing subsequently invalidated the local authentication state. The application then surfaced a synthetic client-side 401, even though the token endpoint had successfully refreshed the session.
I extended the PR to cover this production scenario and the related failure modes:
The renew-running flag is now set before the asynchronous well-known endpoint lookup.
A manual refresh publishes SilentRenewStarted, ensuring that the replayed public event represents the current refresh cycle.
Callers encountering an already-running refresh wait for its NewAuthenticationResult.
SilentRenewFailed is propagated to waiting callers instead of leaving them pending indefinitely.
Waiting callers time out after silentRenewTimeoutInSeconds.
The renew-running flag is reset using finalize, covering success, error, and cancellation.
Regression tests now explicitly cover a periodic silent renew racing with a manual refresh, failure propagation, timeout behavior, and early lock acquisition.
The complete library test suite passes with these changes: 988 tests successful.

ThorstenKunz:
@FabianGosebrink @damienbod any chance to have a look at this? The annoying part of it is, that while I can easily add a single-flight-wrapper around the forceRefreshSession() issue in my consumer to catch the race, but the silentRenew is an internal aspect of the lib so I have no means to fully catch this without disabling silenRenew altogether and reimplement this part entirely in our consumer.

FabianGosebrink (approved):

PR #2187, opened 2026-03-12
Make offline_access refresh-token warning configurable

ThorstenKunz:
## Summary

This change makes the `offline_access` warning for `silentRenew` with `useRefreshToken` explicitly configurable.

The library currently warns whenever refresh tokens are used together with silent renew and the configured scope does not include `offline_access`. That is a reasonable default for many providers, but it is not universally required. Providers such as Keycloak can support this configuration without requiring `offline_access`.

## Changes

- add a new config flag: `disableRefreshTokenOfflineAccessScopeWarning`
- keep the current warning behavior as the default
- suppress the warning when the new flag is enabled
- adjust the warning text so it reads as a provider-dependent recommendation instead of a hard requirement
- add/update unit tests for both default and suppressed warning behavior
- document the new option in the README

## Motivation

This keeps the existing guidance for setups where `offline_access` is expected, while allowing valid provider-specific configurations without producing a misleading console warning.

## Verification

- targeted config validation unit tests pass
- library build passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,12 @@ import { AuthModule, LogLevel } from 'angular-auth-oidc-client';
 export class AppModule {}
 ```
 
+When using `silentRenew` together with `useRefreshToken`, requesting the
+`offline_access` scope is common and remains the default recommendation.
+Some providers support refresh token renewal in the browser without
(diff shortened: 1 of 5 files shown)


FabianGosebrink:
LGTM

@damienbod ?

Issue #2077, opened 2025-03-03
[Question]: Authentication without PKCE - invalid query parameters


andrush001:
**What Version of the library are you using?**
19.0.0

**Question**
If I configure `disablePkce` to `true` I end up un URL: `.../oauth2/authorize?client_id=...&redirect_uri=http%3A%2F%2Flocalhost%3A4200&response_type=code&scope=openid&nonce=c7e0cc46747e9bf7a3e9b44379607aa016BwynIwC&state=3293035c5b0573db792bf437922be409f5lwlwI9g&code_challenge=&code_challenge_method=S256&external=true`

My authentication server throws error that `code_challenge` parameter is invalid. It is indeed invalid, it is empty. I think `code_challenge` and `code_challenge_method` parameter should not be part of the authorization request in case I want to authenticate without PKCE. Am I right?

PR #2209, opened 2026-05-28
Fix: don't send code_challenge in PAR flow when disablePkce is true

FabianGosebrink:
## Summary
- Fixes the `disablePkce` bug from #2077 in the PAR (Pushed Authorization Request) flow path that #2159 didn't cover.
- `createBodyForParCodeFlowRequest` now skips both PKCE param generation and the `code_challenge` / `code_challenge_method` append when `disablePkce: true`.
- Mirrors the existing pattern in `createAuthorizeUrl` by routing challenge generation through the existing private `getCodeChallenge` helper (which short-circuits when PKCE is disabled).

## Background
Issue #2077 reported that strict authorization servers reject `code_challenge=` (empty) when `disablePkce: true`. PR #2159 fixed this for the standard authorize URL by adding a `!configuration.disablePkce` guard in `createAuthorizeUrl`. The PAR flow constructs its request body in a separate method (`createBodyForParCodeFlowRequest`) that was missed — it unconditionally generated a code verifier and appended a populated `code_challenge`, defeating the purpose of `disablePkce` for PAR consumers.

## Changes
- `url.service.ts`: `createBodyForParCodeFlowRequest` now calls the existing `getCodeChallenge(configuration)` helper instead of inlining the verifier+challenge generation, and guards the two `params.append('code_challenge'...)` / `params.append('code_challenge_method'...)` calls behind `!configuration.disablePkce`.
- `url.service.spec.ts`: new test `omits code_challenge and code_challenge_method when disablePkce is true` asserting the output URL is clean and that `createCodeVerifier` / `generateCodeChallenge` are not called.

## Test plan
- [x] New unit test fails before the fix, passes after (verified locally — TDD)
- [x] Full library test suite: 984 / 984 SUCCESS
- [x] \`npm run lint-lib\` clean
- [ ] Manually verify against a real PAR-enabled IdP with \`disablePkce: true\` (recommend a reviewer with a PAR setup confirms)

Refs #2077
Related: #2159

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -619,6 +619,7 @@
       "integrity": "sha512-moERVCTekQKOvR8RMuEOtWSO3VS1qrzA3keI1dPto/JVB8Nqp9w3R5ZpEoXHzh4zgEryosxmPgdi6UczJe2ouQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@angular-eslint/bundled-angular-compiler": "21.3.1",
         "eslint-scope": "9.1.2"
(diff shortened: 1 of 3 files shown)


FabianGosebrink:
@damienbod Please Review.

PR #2211, opened 2026-05-28
fix: verify nonce against original after refresh instead of overwriting it

### LINKED ISSUE #1947: [Bug]:
### Version

17.0.0

_No response_

```true
I'm getting the following error: 

authentication.service.ts:42 [ERROR] 0-ace_fe - silent renew failed! Error: Error: authorizedCallback, token(s) validation failed, resetting. Hash: at angular-auth-oidc-client.mjs:3657:37 at Observable.init [as _subscribe] (throwError.js:5:51) at Observable._trySubscribe (Observable.js:37:25) at Observable.js:31:30 at errorContext (errorContext.js:19:9) at Observable.subscribe (Observable.js:22:21) at catchError.js:14:31 at OperatorSubscriber._error (OperatorSubscriber.js:23:21) at OperatorSubscriber.error (Subscriber.js:40:18) at OperatorSubscriber._error (Subscriber.js:64:30) 
```
```


```true
Angular 17.
"angular-auth-oidc-client": "17.0.0"

Service configuration is set in app.module.ts:

AuthModule.forRoot({ 

config: { 

authority: environment.authenticationConfig.authority, 

redirectUrl: environment.authenticationConfig.redirectUrl, // goes to /redirect 

postLoginRoute: "/home", 

postLogoutRedirectUri: environment.authenticationConfig.postLogoutRedirectUrl,  // goes to /front-page

unauthorizedRoute: environment.authenticationConfig.unauthorizedRoute, // goes to /unauthorized

clientId: "fe", 

ignoreNonceAfterRefresh: false, // if set to true, refresh works but nonce is not validated

scope: "openid profile offline_access", 

responseType: "code", 

silentRenew: true, 

useRefreshToken: true, 

secureRoutes: environment.authenticationConfig.secureRoutes, 

logLevel: LogLevel.Debug, 
```
Authority is a Keycloak container version 23.0.4
FE client configuration: 
```
Client authentication off
Authentication flow: standard flow
```
```


```true
After configuring the authentication service to run an authentication code flow with pkce and silent renew based on refresh tokens, I expected the refreshed tokens nonce validation to be successfull. However the nonce validation raised an error due to the local stage being unconsistent.
```

### Additional context

Authenticate request body:
´´´
client_id: fe 

redirect_uri: http://localhost:4200/redirect

response_type: code 

scope: openid profile offline_access 

nonce: 073d4fc4245855ebcc11fb1aaf2bf6ec8fvf3PHd6 

state: ac41b608d9eb7b6a12f92803596728ebdfQu9Jnek 

code_challenge: m-67CbyWPkVs0NtMOEFXsDiO8MpkF1klNEaiU4QCTKY 

code_challenge_method: S256 
´´´

Information stored in local storage after redirection by keycloak:
```
access_token_expires_at: 1716886568000 

authNonce: null 

authStateControl: "" 

authWellKnownEndPoints: {issuer: "http://localhost:8080/realms/myapp",…} 

authnResult: {,…} 

authzData: "eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICI3eVJHRVRBVXctdTNCWnVFTDF1WW5jWkR3eURwXzlYTnJ6djctMUFGVV8wIn0.eyJleHAiOjE3MTY4ODY1NjgsImlhdCI6MTcxNjg4NjI2OCwiYXV0aF90aW1lIjoxNzE2ODg2MjY3LCJqdGkiOiJjMjkxZjVjNS1hMTkwLTQ0ODEtODZmMS0wMGEwZmJlYjY0YzYiLCJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjgwODAvcmVhbG1zL2FjZSIsImF1ZCI6ImFjY291bnQiLCJzdWIiOiIwMWE1NjIwYy1iNzFhLTRiYjQtYmI2MC01ZjAzMDM2NzI0MDUiLCJ0eXAiOiJCZWFyZXIiLCJhenAiOiJhY2VfZmUiLCJub25jZSI6IjA3M2Q0ZmM0MjQ1ODU1ZWJjYzExZmIxYWFmMmJmNmVjOGZ2ZjNQSGQ2Iiwic2Vzc2lvbl9zdGF0ZSI6ImU1YTA2Y2EwLTRmMmMtNDE4ZC04ZDQzLTg1ZWQyOTg2MGJlZiIsImFjciI6IjEiLCJhbGxvd2VkLW9yaWdpbnMiOlsiaHR0cDovL2xvY2FsaG9zdDo0MjAwIl0sInJlYWxtX2FjY2VzcyI6eyJyb2xlcyI6WyJwcmltYXJ5X2NhcmUiLCJkZWZhdWx0LXJvbGVzLXRlc3QiLCJvZmZsaW5lX2FjY2VzcyIsInVtYV9hdXRob3JpemF0aW9uIl19LCJyZXNvdXJjZV9hY2Nlc3MiOnsiYWNjb3VudCI6eyJyb2xlcyI6WyJtYW5hZ2UtYWNjb3VudCIsIm1hbmFnZS1hY2NvdW50LWxpbmtzIiwidmlldy1wcm9maWxlIl19fSwic2NvcGUiOiJvcGVuaWQgcHJvZmlsZSBvZmZsaW5lX2FjY2VzcyBlbWFpbCIsInNpZCI6ImU1YTA2Y2EwLTRmMmMtNDE4ZC04ZDQzLTg1ZWQyOTg2MGJlZiIsImVtYWlsX3ZlcmlmaWVkIjpmYWxzZSwicHJlZmVycmVkX3VzZXJuYW1lIjoicGNhcmUifQ.ENZr-_NqvE9tQVTqWIZt4dDSzvcd00ZbdFKlEMSqi2SAnkhaLhioFvzpUefp6BZIAIPuiAk6P7yiPPmlUjFDtEzC4x0jz6201J9n1g1X645dS7MkX6ePRY3dhe56urXefBds3FyU7x6rhOLK1r6LJALoJL3NOLtXvMoTSyQo49ZcQd2DDFTGWhZ7Ktj8Fgp5UWTOCn7H0EFJV4m3MU_EUKJz6uKejBeLeudLZWTnhZ4tqPVwi15jd4CvZ_xrlXJHkhfsyU6UKlvOKf8Tb0htZQyXttGu-VxTWL8sPPexR6upsvOQMw_XAnXjiRb0EEL5N1GLxJbRidu-O7_ceUPq3w" 

codeVerifier: "581a94aa63e6c12d6fff5cd17e1c4c8208a434d8c75f496f2477c787c317Y2VZ5iU" 

jwtKeys: {,…} 

session_state: "e5a06ca0-4f2c-418d-8d43-85ed29860bef" 

storageCodeFlowInProgress: false 

storageSilentRenewRunning: "" 
```
We can see that the authNonce variable is null. I think this is causing the error. Maybe thie original nonce is being overwritten by some of the calls to the setNonce function, for example the one in the src/lib/flows/callback-handling/history-jwt-keys-callback-handler.service.ts file, or the one in the src/lib/iframe/silent-renew.service.ts which could mean that maybe something related to the keys or to the params of the renew callback are wrong?

I am also getting, this log (not really an error?) which I think is the cause of the token validation error:
```
Validate_id_token_nonce failed, dataIdToken.nonce: 073d4fc4245855ebcc11fb1aaf2bf6ec8fvf3PHd6 local_nonce:--RefreshToken-- 
``` 
It is getting the nonce from the token buth the stored nonce is a placeholder.
This placeholder is used in the file src/lib/flows/callback-handling/refresh-session-callback-handler.service.ts to set the authNonce state to the placeholder if there is refresh token in the state service?

angular-auth-oidc-client/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/refresh-session-callback-handler.service.ts
```
// Nonce is not used with refresh tokens; but Key cloak may send it anyway 

this.flowsDataService.setNonce( 

TokenValidationService.refreshTokenNoncePlaceholder, 

config 

); 
```
Indeed my Keycloak instance sends the nonce in every token of the response to the refresh token request.   

The localNonce may be later check in the oidc-client/src/lib/validation/token-validation.service.ts file where the failed check log message that troubles me is raised:
´´´ 
if (!isFromRefreshToken && dataIdToken.nonce !== localNonce) { 

this.loggerService.logDebug( 

configuration, 

'Validate_id_token_nonce failed, dataIdToken.nonce: ' + 

dataIdToken.nonce + 

' local_nonce:' + 

localNonce 

); 

 

return false; 

} 
```

¿Maybe is this check that is causing the token validation to fail?

This are the other warnings and errors that I get from the console:
```
authentication.service.ts:42 [WARN] 0-ace_fe - authCallback incorrect nonce, did you call the checkAuth() method multiple times? 

angular-auth-oidc-client.mjs:70 [DEBUG] 0-ace_fe - authCallback token(s) invalid 

authentication.service.ts:42 [WARN] 0-ace_fe - authorizedCallback, token(s) validation failed, resetting. Hash:  

angular-auth-oidc-client.mjs:70 [DEBUG] 0-ace_fe - Local Login information cleaned up and event fired 

angular-auth-oidc-client.mjs:70 [DEBUG] 0-ace_fe - Local Login information cleaned up and event fired 

 

authentication.service.ts:42 [ERROR] 0-ace_fe - silent renew failed! Error: Error: authorizedCallback, token(s) validation failed, resetting. Hash: at angular-auth-oidc-client.mjs:3657:37 at Observable.init [as _subscribe] (throwError.js:5:51) at Observable._trySubscribe (Observable.js:37:25) at Observable.js:31:30 at errorContext (errorContext.js:19:9) at Observable.subscribe (Observable.js:22:21) at catchError.js:14:31 at OperatorSubscriber._error (OperatorSubscriber.js:23:21) at OperatorSubscriber.error (Subscriber.js:40:18) at OperatorSubscriber._error (Subscriber.js:64:30) 
```

### LINKED ISSUE #2068: [Question]: Nonce is not generated for silentRefresh
18.0.2
code workflow

**Question**
Hey, we were investigating a problem in our application where the silent refresh isn't working due to a nonce check. We came across various issues where the suggested solution was to deactivate the nonce check using:

`ignoreNonceAfterRefresh: true,`

However, we were not comfortable with disabling this check and decided to investigate why this issue was occurring. During our investigation, we found the following lines of code in the silent refresh functionality:

RefreshSessionCallbackHandlerService.ts
```
static { this.refreshTokenNoncePlaceholder = '--RefreshToken--';}

// Nonce is not used with refresh tokens; but Key cloak may send it anyway
this.flowsDataService.setNonce(TokenValidationService.refreshTokenNoncePlaceholder, config);
```

Our understanding is that every time the silent refresh runs, it sends this placeholder as the nonce instead of generating a new one or the inital one. When the response is received, the system checks for the nonce from the request against this placeholder string. This check fails because it seems like our SSO server does not accept the string from refreshTokenNoncePlaceholder and instead returns the nonce which was used for authenfication in the first place.

I managed to inject my initial nonce, which was generated before, into a silent renewal request and the nonce check worked as expected.

I'm not deeply familiar with how the OIDC and OAuth workflows work, but from my understanding, this could also be a potential attack vector. 
Apologies if I am missing something in the authentication workflow.

### PR DESCRIPTION
**Pull request by FabianGosebrink**
## Summary
- Stop writing the placeholder string `'--RefreshToken--'` over the stored `authNonce` during silent renew.
- Keep the original nonce in storage and pass an explicit `isRefreshTokenFlow` flag into `TokenValidationService.validateIdTokenNonce`.
- New refresh-flow branch verifies that any returned nonce equals the original — so the Keycloak case works without needing to disable the check.

Fixes #1947
Fixes #2068

## Background

Per OIDC core spec section 12.2, the id_token from a refresh-token exchange SHOULD NOT contain a `nonce` claim. Some IdPs (notably **Keycloak**, also reported with other providers) include the original nonce anyway. Before this PR, the library:

1. Overwrote `authNonce` in storage with the literal `'--RefreshToken--'` before each refresh (`refresh-session-callback-handler.service.ts`).
2. Then compared the IdP's returned nonce against that placeholder.
3. → mismatch → silent renew fails → user logged out.

The only escape was `ignoreNonceAfterRefresh: true`, which **disables the nonce check entirely** post-refresh. Several users on #1947 / #2068 explicitly rejected that workaround for security reasons.

## New validation matrix (refresh-token flow)

| `id_token.nonce` | Matches original? | `ignoreNonceAfterRefresh` | Result |
|---|---|---|---|
| `undefined` | — | — | ✅ accept (spec-compliant IdP) |
| defined | yes | — | ✅ accept (Keycloak — verified) |
| defined | no | `true` | ✅ accept (preserved escape hatch) |
| defined | no | `false` | ❌ reject (real anomaly — was previously hidden by the placeholder) |

Normal (non-refresh) auth-flow behavior is unchanged.

This is **stricter** than `ignoreNonceAfterRefresh: true`: Keycloak users now get nonce *verification* against the original instead of nonce *skipping*, so tampering with the returned nonce is still caught.

## API change

`TokenValidationService.validateIdTokenNonce` gained an optional 5th parameter:

\`\`\`ts
validateIdTokenNonce(
  dataIdToken,
  localNonce,
  ignoreNonceAfterRefresh,
  configuration,
  isRefreshTokenFlow = false,   // NEW, defaults to false
): boolean
\`\`\`

Default is `false`, so existing 4-arg callers behave identically. The library's own call site in `state-validation.service.ts:138` passes the `isInRefreshTokenFlow` value already computed at line 85.

The static `TokenValidationService.refreshTokenNoncePlaceholder` is **kept and marked `@deprecated`** — no longer written to storage, but the symbol stays exported so any external consumer that imported it still compiles. Plan to remove in the next major.

## Migration note

Users mid-session at upgrade time who still have `'--RefreshToken--'` in localStorage from a previous version will see one more cycle of the bug on their *next* refresh (because the stored nonce isn't the real one). After the next successful login the real nonce is written and the issue clears. No user action needed.

## Test plan
- [x] 4 rewritten `validateIdTokenNonce` tests + 1 new tampered-nonce test — all pass
- [x] New `refresh-session-callback-handler` test asserts `setNonce` is not called when a refresh token exists
- [x] Full library test suite: **984 / 984 SUCCESS**
- [x] \`npm run lint-lib\` clean
- [ ] Manually verify against a real Keycloak with default `ignoreNonceAfterRefresh: false` — recommend a reviewer with a Keycloak setup confirms

## Files changed
- \`validation/token-validation.service.ts\` — signature + new logic + @deprecated tag
- \`validation/token-validation.service.spec.ts\` — rewrote 4 refresh-flow tests, added tamper-rejection
- \`validation/state-validation.service.ts\` — pass `isInRefreshTokenFlow` at the call site
- \`flows/callback-handling/refresh-session-callback-handler.service.ts\` — drop placeholder overwrite
- \`flows/callback-handling/refresh-session-callback-handler.service.spec.ts\` — assert no `setNonce` call

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/validation/token-validation.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/validation/token-validation.service.ts
@@ -56,6 +56,11 @@ import { alg2kty, getImportAlg, getVerifyAlg } from './token-validation.helper';
 
 @Injectable({ providedIn: 'root' })
 export class TokenValidationService {
+  /**
+   * @deprecated No longer written to storage. Kept only so external
+   * consumers that previously imported the symbol still compile.
+   * Will be removed in a future major release.
+   */
   static refreshTokenNoncePlaceholder = '--RefreshToken--';
 
   keyAlgorithms: string[] = [
@@ -282,13 +287,34 @@ export class TokenValidationService {
     dataIdToken: any,
     localNonce: any,
     ignoreNonceAfterRefresh: boolean,
-    configuration: OpenIdConfiguration
+    configuration: OpenIdConfiguration,
+    isRefreshTokenFlow = false
   ): boolean {
-    const isFromRefreshToken =
-      (dataIdToken.nonce === undefined || ignoreNonceAfterRefresh) &&
-      localNonce === TokenValidationService.refreshTokenNoncePlaceholder;
+    if (isRefreshTokenFlow) {
+      if (dataIdToken.nonce === undefined) {
+        return true;
+      }
 
-    if (!isFromRefreshToken && dataIdToken.nonce !== localNonce) {
+      if (ignoreNonceAfterRefresh) {
+        return true;
+      }
+
+      if (dataIdToken.nonce === localNonce) {
+        return true;
+      }
+
+      this.loggerService.logDebug(
+        configuration,
+        'Validate_id_token_nonce failed in refresh-token flow, dataIdToken.nonce: ' +
+          dataIdToken.nonce +
+          ' local_nonce:' +
+          localNonce
+      );
+
+      return false;
+    }
+
+    if (dataIdToken.nonce !== localNonce) {
       this.loggerService.logDebug(
         configuration,
         'Validate_id_token_nonce failed, dataIdToken.nonce: ' +
--- a/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/refresh-session-callback-handler.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/refresh-session-callback-handler.service.spec.ts
@@ -78,5 +78,23 @@ describe('RefreshSessionCallbackHandlerService', () => {
           },
         });
     }));
+
+    it('does not overwrite the stored nonce when a refresh token exists', waitForAsync(() => {
+      spyOn(
+        flowsDataService,
+        'getExistingOrCreateAuthStateControl'
+      ).and.returnValue('state-data');
+      spyOn(authStateService, 'getRefreshToken').and.returnValue(
+        'henlo-furiend'
+      );
+      spyOn(authStateService, 'getIdToken').and.returnValue('henlo-legger');
+      const setNonceSpy = spyOn(flowsDataService, 'setNonce');
+
+      service
+        .refreshSessionWithRefreshTokens({ configId: 'configId1' })
+        .subscribe(() => {
+          expect(setNonceSpy).not.toHaveBeenCalled();
+        });
+    }));
   });
 });
--- a/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/refresh-session-callback-handler.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/flows/callback-handling/refresh-session-callback-handler.service.ts
@@ -3,7 +3,6 @@ import { Observable, of, throwError } from 'rxjs';
 import { AuthStateService } from '../../auth-state/auth-state.service';
 import { OpenIdConfiguration } from '../../config/openid-configuration';
 import { LoggerService } from '../../logging/logger.service';
-import { TokenValidationService } from '../../validation/token-validation.service';
 import { CallbackContext } from '../callback-context';
 import { FlowsDataService } from '../flows-data.service';
 
@@ -43,11 +42,6 @@ export class RefreshSessionCallbackHandlerService {
       this.loggerService.logDebug(
         config,
         'found refresh code, obtaining new credentials with refresh code'
-      );
-      // Nonce is not used with refresh tokens; but Key cloak may send it anyway
-      this.flowsDataService.setNonce(
-        TokenValidationService.refreshTokenNoncePlaceholder,
-        config
       );
 
       return of(callbackContext);
--- a/projects/angular-auth-oidc-client/src/lib/validation/state-validation.service.ts
+++ b/projects/angular-auth-oidc-client/src/lib/validation/state-validation.service.ts
@@ -139,7 +139,8 @@ export class StateValidationService {
                 toReturn.decodedIdToken,
                 authNonce,
                 Boolean(ignoreNonceAfterRefresh),
-                configuration
+                configuration,
+                isInRefreshTokenFlow
               )
             ) {
               this.loggerService.logWarning(
--- a/projects/angular-auth-oidc-client/src/lib/validation/token-validation.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/validation/token-validation.service.spec.ts
@@ -111,56 +111,52 @@ describe('TokenValidationService', () => {
       ).toBe(false);
     });
 
-    it('should validate id token nonce after refresh token grant when undefined and no ignore', () => {
+    it('validates refresh-token flow when id_token contains no nonce (spec-compliant IdP)', () => {
       expect(
         tokenValidationService.validateIdTokenNonce(
           { nonce: undefined },
-          TokenValidationService.refreshTokenNoncePlaceholder,
+          'realNonce',
           false,
-          {
-            configId: 'configId1',
-          }
+          { configId: 'configId1' },
+          true
         )
       ).toBe(true);
     });
 
-    it('should validate id token nonce after refresh token grant when undefined and ignore', () => {
+    it('validates refresh-token flow when id_token nonce matches the original nonce (Keycloak case)', () => {
       expect(
         tokenValidationService.validateIdTokenNonce(
-          { nonce: undefined },
-          TokenValidationService.refreshTokenNoncePlaceholder,
-          true,
-          {
-            configId: 'configId1',
-          }
-        )
-      ).toBe(true);
-    });
-
-    it('should validate id token nonce after refresh token grant when defined and ignore', () => {
-      expect(
-        tokenValidationService.validateIdTokenNonce(
-          { nonce: 'test1' },
-          TokenValidationService.refreshTokenNoncePlaceholder,
-          true,
-          {
-            configId: 'configId1',
-          }
-        )
-      ).toBe(true);
-    });
-
-    it('should not validate id token nonce after refresh token grant when defined and no ignore', () => {
-      expect(
-        tokenValidationService.validateIdTokenNonce(
-          { nonce: 'test1' },
-          TokenValidationService.refreshTokenNoncePlaceholder,
+          { nonce: 'realNonce' },
+          'realNonce',
           false,
-          {
-            configId: 'configId1',
-          }
+          { configId: 'configId1' },
+          true
+        )
+      ).toBe(true);
+    });
+
+    it('rejects refresh-token flow when id_token nonce differs from original and ignoreNonceAfterRefresh is false', () => {
+      expect(
+        tokenValidationService.validateIdTokenNonce(
+          { nonce: 'tamperedNonce' },
+          'realNonce',
+          false,
+          { configId: 'configId1' },
+          true
         )
       ).toBe(false);
+    });
+
+    it('validates refresh-token flow when ignoreNonceAfterRefresh is true regardless of returned nonce', () => {
+      expect(
+        tokenValidationService.validateIdTokenNonce(
+          { nonce: 'anythingFromIdP' },
+          'realNonce',
+          true,
+          { configId: 'configId1' },
+          true
+        )
+      ).toBe(true);
     });
   });
 

### CODE REVIEW

File: projects/angular-auth-oidc-client/src/lib/validation/token-validation.service.ts [hunk 1]
```diff
@@ -56,6 +56,11 @@ import { alg2kty, getImportAlg, getVerifyAlg } from './token-validation.helper';
 
 @Injectable({ providedIn: 'root' })
 export class TokenValidationService {
+  /**
+   * @deprecated No longer written to storage. Kept only so external
+   * consumers that previously imported the symbol still compile.
+   * Will be removed in a future major release.
+   */
   static refreshTokenNoncePlaceholder = '--RefreshToken--';
```
damienbod, line 65: Why is this not stored on the storage, what happens after a refresh?

Not saying this is bad, just want to understand this better. Also wondering if this is a behavior change

FabianGosebrink: All Good. It is good that you double check this.

In the old code the RefreshSessionCallbackHandlerService overwrote authNonce in storage with '--RefreshToken--' before each refresh. validateIdTokenNonce then detected refresh flow by checking if localNonce === refreshTokenNoncePlaceholder. So the storage was being used as a flag and the real initial nonce was getting overwritten.

Now it is not stored in the storage anymore. authNonce in storage now holds the original nonce from the initial authentication, for the whole session. It's never overwritten during refresh. state-validation.service.ts already computes isInRefreshTokenFlow at line 85 from the callback context (isRenewProcess && !!refreshToken), and that boolean is now passed directly into validateIdTokenNonce instead of being used via storage.

It is a change, yes.

| IdP returns in the refreshed `id_token` | Old behavior | New behavior |
|---|---|---|
| No `nonce` (spec-compliant) | ✅ accepted | ✅ accepted |
| Original `nonce` (Keycloak) | ❌ rejected (placeholder ≠ returned nonce) | ✅ accepted (returned nonce == original in storage) |
| Some other `nonce` | ❌ rejected | ❌ rejected |
| `ignoreNonceAfterRefresh: true` | ✅ accepted | ✅ accepted (unchanged) |

So yes — **this is a behavior change**, and it's intentional. Two improvements:

1. **Keycloak (and other IdPs that echo the nonce on refresh) now work with the default `ignoreNonceAfterRefresh: false`.** Previously they required users to flip the flag, which weakens the check.
2. **Tampering is now caught.** If a man-in-the-middle returns a nonce that *isn't* the original, validation fails. The old code couldn't tell the difference between "spec-compliant IdP" and "tampered refresh" — both produced a mismatch with the placeholder.

Net security posture is **stricter**, not looser. `ignoreNonceAfterRefresh: true` still exists as an escape hatch for IdPs that return arbitrary nonces.

**On the deprecated static**

The `refreshTokenNoncePlaceholder` constant is still exported (just unused internally) so any external consumer that imported it still compiles. Plan is to drop it in the next major. Happy to remove it now instead if you'd prefer a cleaner break — it's only useful as a transition aid.

PR #2215, opened 2026-05-28
test: convert validation specs to await firstValueFrom

FabianGosebrink:
## Summary
Replace the `waitForAsync(() => observable.subscribe(cb))` pattern with `async () => { const x = await firstValueFrom(observable); ... }` across the three validation specs.

**3 files, 28 tests converted**, all pass; full lib suite still green; lint clean.

## Why
The old pattern silent-passes when the observable never emits. If a regression in production code, a bad mock, or a missing pipe operator stops emission, the `subscribe` callback never runs, **no `expect()` ever fires**, and `waitForAsync` is satisfied because the zone has no pending tasks. The test "passes" while exercising nothing.

With `await firstValueFrom`, a never-emit hangs the promise and Jasmine times out the test. A regression that breaks the path under test now shows up as a real failure instead of being silently masked.

This matters most in `validation/` — these specs guard token/signature/state checks, the security core of the library. A regression slipping past these tests is exactly the kind of bug you don't want hidden.

## What's converted

| File | Tests converted | Notes |
|---|---|---|
| `jwt-window-crypto.service.spec.ts` | 1 | Single happy-path test in `generateCodeChallenge` |
| `token-validation.service.spec.ts` | 5 | All `validateSignatureIdToken` tests using the silent-pass pattern. The `done`-callback tests in this file are already safe — left untouched. |
| `state-validation.service.spec.ts` | 22 | All `getValidatedStateResult` tests |

## Two commits

1. **`chore: normalize line endings on two validation specs`** — `token-validation.service.spec.ts` and `state-validation.service.spec.ts` were checked in with mixed CRLF/CR/LF line endings, which git's auto-detection treated as binary (`-text`). Re-encoded to LF to match `core.autocrlf = input`. **No semantic change** — review this commit in seconds by scanning for `-text` vs `lf` in `git ls-files --eol`.
2. **`test: convert validation specs to await firstValueFrom`** — the actual conversions. ~325 / -330 lines, all in the 3 files listed above.

Splitting the line-ending normalization into its own commit keeps the substantive diff readable.

## Test plan
- [x] All 28 converted tests pass
- [x] Full library test suite passes (~985 tests)
- [x] Sanity-checked the new pattern by deliberately introducing a wrong expected value — the converted test failed as expected (proving silent-pass is gone)

## Out of scope (intentional)
- Same pattern exists in `callback/` and `flows/` specs (~17 more files, ~90 sites). To keep this PR reviewable, those will be addressed in follow-up PRs.
- The `done`-callback tests already in the converted files are already safe and were not touched.

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/validation/jwt-window-crypto.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/validation/jwt-window-crypto.service.spec.ts
@@ -1,4 +1,5 @@
-import { TestBed, waitForAsync } from '@angular/core/testing';
+import { TestBed } from '@angular/core/testing';
+import { firstValueFrom } from 'rxjs';
 import { CryptoService } from '../utils/crypto/crypto.service';
 import { JwtWindowCryptoService } from './jwt-window-crypto.service';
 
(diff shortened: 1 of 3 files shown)


FabianGosebrink:
@damienbod

PR #2220, opened 2026-05-29
fix: strip hash fragment when comparing redirect vs current URL (closes #2143)

### LINKED ISSUE #2143: [Bug]: Facebook appending #_=_ to redirected url
### Version

20.0.2

_No response_

```true
No action after facebook login. Page stuck in landing url.
Url after redirect from facebook authorize url is something like:

https://example.com/auth-callback?state=...&iss=...&code=...#_=_

More info about this facebook behavior: https://stackoverflow.com/a/41917323/3791820

Also I found question on SO with the same problem: https://stackoverflow.com/questions/79310174/angular-auth-oidc-client-with-facebook
```

```true

```

```true
I think in UrlService you should compare redirectUrl and  currentUrl without QueryParams AND fragment
```

### Additional context

_No response_

FabianGosebrink:
## Summary
`UrlService.getUrlWithoutQueryParameters` now also clears `u.hash`. That's the entire production change — 1 line + a comment. Closes #2143 (Facebook `#_=_` callback stuck) and fixes a related implicit-flow regression two commenters reported on the same issue.

**Two commits** so the diff is reviewable:
1. `chore: normalize line endings on url.service.spec.ts` — the spec file was checked in with mixed CRLF/CR/LF (git treated it as `-text`/binary). Re-encode to LF to match the rest of the repo. No semantic change.
2. `fix: strip hash fragment when comparing redirect vs current URL` — the actual fix + 3 new tests.

## The bug

With `checkRedirectUrlWhenCheckingIfIsCallback: true`, `isCallbackFromSts` compared `currentUrl` and `redirectUrl` after stripping the query string. The hash fragment was kept. Two providers were silently broken:

- **Facebook** appends `"#_=_"` to OAuth redirects ([known behavior](https://stackoverflow.com/a/41917323/3791820)). Result: `https://app.com/cb?code=…#_=_` was rejected because `https://app.com/cb#_=_` ≠ `https://app.com/cb`.
- **Implicit flow** puts the auth response in the fragment by design (`#access_token=…&state=…`). Same root cause: the fragment made the URL strings unequal.

Both reported in #2143 — original author for Facebook, two follow-up commenters for implicit flow and a regression on the v17→v20 upgrade.

## The fix

```diff
 keys.forEach((key) => {
   u.searchParams.delete(key);
 });

+// Also drop the hash fragment. Some identity providers (e.g. Facebook)
+// append "#_=_" to their redirect URL, and implicit flow puts the auth
+// response in the fragment by design. Either way, the fragment is not
+// part of the configured redirect URL and must not factor into the
+// base-URL comparison done by isCallbackFromSts.
+// https://github.com/damienbod/angular-auth-oidc-client/issues/2143
+u.hash = '';

 return u;
```

The companion check (the configured redirect's query parameters must all be present on the current URL) is **unchanged** — it still defends against partial-match callback URLs.

## Does it change behaviour?

**Yes — but only in the direction the bug report demands.**

- `isCallbackFromSts` may now return `true` in cases where it previously (incorrectly) returned `false`: specifically, when `currentUrl` carries a hash fragment that the configured `redirectUrl` doesn't have. That covers the Facebook + implicit-flow cases above.
- It does **not** turn any previously-`true` result into `false`. Nothing that worked before stops working.
- The query-parameter presence check is untouched, so a current URL missing the redirect URL's required query params still fails as before.

The only edge case would be a user who **configures** `redirectUrl` with a hash fragment (e.g. `https://app.com/cb#foo`) and expects the fragment to be enforced. That's against the OAuth spec ([RFC 6749 §3.1.2](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2): redirect URI MUST NOT include a fragment) and would have been broken in other ways even before this PR.

Internal helper note: `getUrlWithoutQueryParameters` is now slightly misnamed (it also strips the fragment). The method isn't exported via the lib's `index.ts` so the rename can happen separately if you want. Left as-is here to keep the fix minimal.

## Test plan
- [x] 3 new tests, all failed before the fix:
  - helper: `getUrlWithoutQueryParameters` also strips the hash
  - `isCallbackFromSts`: Facebook `#_=_` regression
  - `isCallbackFromSts`: implicit-flow fragment
- [x] Full library test suite passes (101/101 in this spec; full suite green)

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/utils/url/url.service.spec.ts
@@ -45,7 +45,8 @@ describe('UrlService Tests', () => {
 
   describe('getUrlWithoutQueryParameters', () => {
     it('should return a new instance of the passed URL without any query parameters', () => {
-      const url = new URL('https://any.url');      const params = [
+      const url = new URL('https://any.url');
+      const params = [
         { key: 'doot', value: 'boop' },
         { key: 'blep', value: 'blep' },
       ];
@@ -63,7 +64,8 @@ describe('UrlService Tests', () => {
   });
 
   describe('queryParametersExist', () => {
-    const expected = new URLSearchParams();    const params = [
+    const expected = new URLSearchParams();
+    const params = [
       { key: 'doot', value: 'boop' },
       { key: 'blep', value: 'blep' },
     ];
@@ -75,7 +77,8 @@ describe('UrlService Tests', () => {
     const matchingUrls = [
       new URL('https://any.url?doot=boop&blep=blep'),
       new URL('https://any.url?doot=boop&blep=blep&woop=doot'),
-    ];    const nonMatchingUrls = [
+    ];
+    const nonMatchingUrls = [
       new URL('https://any.url?doot=boop'),
       new URL('https://any.url?blep=blep&woop=doot'),
     ];
@@ -351,7 +354,8 @@ describe('UrlService Tests', () => {
         'nonce',
         'state',
         config
-      );      const expectValue =
+      );
+      const expectValue =
         'http://example?client_id=188968487735-b1hh7k87nkkh6vv84548sinju2kpr7gn.apps.googleusercontent.com' +
         '&redirect_uri=https%3A%2F%2Flocalhost%3A44386' +
         '&response_type=code' +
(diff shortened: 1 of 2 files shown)


damienbod:
This looks like a previous PR already in the repo. I need to verify, if the causes new bugs

PR #2224, opened 2026-05-30
fix: reset abandoned code flow flag on non-callback checkAuth (#2221)

### LINKED ISSUE #2221: [Bug]: storageCodeFlowInProgress flag never reset when checkAuthIncludingServer establishes session via SSO silent renew after an abandoned code flow
### Version

21.0.0

_No response_

```true
No exception is thrown. The periodic token expiry check silently stops working. After a successful checkAuthIncludingServer call (which authenticates via iframe silent renew using an existing SSO session), the shouldStartPeriodicallyCheckForConfig method always returns false because isCodeFlowInProgress reads true from storage indefinitely. Tokens are never silently renewed after that point.
```

```true
1. Configure the library with silentRenew: true and useRefreshToken: false (iframe-based silent renew).
2. Call authorize() — the user is redirected to the Identity Provider. storageCodeFlowInProgress is set to true.
3. The user authenticates at the IdP (SSO session is established) but does not return via the expected authorization callback URL (e.g. navigates directly back to the app root).
4. The app calls checkAuthIncludingServer(). Since there are no local tokens, checkAuthWithConfig returns isAuthenticated: false.
5. forceRefreshSession is called and successfully retrieves tokens via iframe silent renew (SSO session exists at the IdP).
6. The user is now authenticated with valid tokens in storage.
7. Observe that startTokenValidationPeriodically never triggers a token refresh — shouldStartPeriodicallyCheckForConfig always returns false.
```

```true
After checkAuthIncludingServer successfully authenticates the user via silent renew, the periodic token check should work normally and silently renew tokens before they expire, just as it would after a regular checkAuth callback flow.
```

### Additional context

The root cause is that `resetCodeFlowInProgress` is only called inside `CodeFlowCallbackService.authenticatedCallbackWithCode` (code-flow-callback.service.ts:32), which is only reached when the current URL contains an authorization code (i.e. [isCallback === true]). When the code flow is abandoned and `checkAuthIncludingServer`takes the `forceRefreshSession`path instead, `resetCodeFlowInProgress` is never called, leaving storageCodeFlowInProgress = true in storage permanently.

Contrast this with isSilentRenewRunning which has a built-in timeout mechanism (silentRenewTimeoutInSeconds) that self-heals if the process gets stuck. isCodeFlowInProgress has no equivalent safeguard.

Suggested fix — reset the flag in `checkAuthWithConfig` when no callback is detected, since that implies any previously started code flow was abandoned:

```javascript
// check-auth.service.ts — checkAuthWithConfig
const isCallback = this.callbackService.isCallback(currentUrl, config);

if (!isCallback) {
  this.flowsDataService.resetCodeFlowInProgress(config);
}
```
This is the minimal, defensive fix: if the app loads without a callback URL, there is no in-flight code flow to protect, so the flag should always be cleared.

FabianGosebrink:
## Summary

Fixes #2221 — the `storageCodeFlowInProgress` flag was never reset when a code flow is abandoned and the session is later established via silent renew, permanently disabling periodic token renewal.

### Root cause

`resetCodeFlowInProgress` is only called in `CodeFlowCallbackService.authenticatedCallbackWithCode`, which is reached **only** when the current URL contains an authorization code (`isCallback === true`).

When a user calls `authorize()` (which sets `storageCodeFlowInProgress = true`), authenticates at the IdP, but returns to the app **without** the callback URL, the flag is never reset. If the session is then established via `checkAuthIncludingServer` → `forceRefreshSession` (iframe silent renew), `shouldStartPeriodicallyCheckForConfig` keeps reading `isCodeFlowInProgress === true` and always returns `false` — so tokens are silently never renewed.

Unlike `isSilentRenewRunning`, `isCodeFlowInProgress` has no self-healing timeout.

### Fix

Reset the flag in `checkAuthWithConfig` whenever the current URL is **not** a callback. If the app loads without a callback URL, there is no in-flight code flow to protect, so the flag should be cleared. This is the minimal, defensive fix suggested in the issue, and it does not interfere with iframe silent renew (gated by the separate `isSilentRenewRunning` flag) or the existing reset on the callback path.

### Tests

Added two tests to `check-auth.service.spec.ts` (written test-first):
- resets a previously abandoned code flow when the current URL is **not** a callback
- does **not** reset the code flow when the current URL **is** a callback (the existing callback-path reset remains the single source of truth during real flows)

### Verification

- ✅ 995/995 lib tests pass (`npm run test-lib-ci`)
- ✅ `npm run lint-lib` clean
- ✅ Prettier clean

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/auth-state/check-auth.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/auth-state/check-auth.service.spec.ts
@@ -12,6 +12,7 @@ import {
 } from '../config/loader/config-loader';
 import { OpenIdConfiguration } from '../config/openid-configuration';
 import { CallbackContext } from '../flows/callback-context';
+import { FlowsDataService } from '../flows/flows-data.service';
 import { CheckSessionService } from '../iframe/check-session.service';
 import { SilentRenewService } from '../iframe/silent-renew.service';
 import { LoggerService } from '../logging/logger.service';
@@ -39,6 +40,7 @@ describe('CheckAuthService', () => {
   let storagePersistenceService: StoragePersistenceService;
   let currentUrlService: CurrentUrlService;
   let publicEventsService: PublicEventsService;
+  let flowsDataService: FlowsDataService;
 
   beforeEach(() => {
     TestBed.configureTestingModule({
@@ -58,6 +60,7 @@ describe('CheckAuthService', () => {
         mockAbstractProvider(StsConfigLoader, StsConfigStaticLoader),
         AutoLoginService,
         mockProvider(StoragePersistenceService),
+        mockProvider(FlowsDataService),
       ],
     });
   });
@@ -78,6 +81,7 @@ describe('CheckAuthService', () => {
     storagePersistenceService = TestBed.inject(StoragePersistenceService);
     currentUrlService = TestBed.inject(CurrentUrlService);
     publicEventsService = TestBed.inject(PublicEventsService);
+    flowsDataService = TestBed.inject(FlowsDataService);
   });
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


damienbod:
I would need to verify that this has no side effects

Issue #1662, opened 2023-01-26
[Bug]: Silent refresh fails using localStorage when multiple tabs trying to refresh at the same time


medeirosrich:
### Version

5.0.3

### Please provide the exception or error you saw

```true
silent renew failed! Error: Error: authorizedCallback, token(s) validation failed, resetting.
```

### Steps to reproduce the behavior

```true
1) Set up OIDC Code Flow PKCE using refresh tokens (used 'ng add angular-auth-oidc-client')

2) Set up custom storage using localStorage (used example from https://angular-auth-oidc-client.com/docs/documentation/custom-storage)

4) To speed up testing, setup your authority to have short lived access and refresh tokens. (i.e. 1 or 2 minutes) 

3) Login using one browser [chrome] tab (i.e. this.oidcSecurityService.authorize()). Open console and verify refresh is working

4) Open a few more tabs. Open console and verify they are authorized and refreshing

5) Wait until error occurs. It shouldn't take long - usually 2-10 minutes with 1 minute tokens. Open more tabs if needed.
```

### A clear and concise description of what you expected to happen.

```true
access token should refresh successfully and all tabs remain authorized
```

### Additional context

It seems that multiple tabs simultaneously attempting to refresh the token using a shared localStorage is causing a race condition, resulting in validation failure.

TarasKovalenko:
@medeirosrich I tried to reproduce the issue, but from my side, everything works well. Can you please prove the AuthModule configuration?

medeirosrich:
> @medeirosrich I tried to reproduce the issue, but from my side, everything works well. Can you please prove the AuthModule configuration?
```
config: {
            authority: '[my authority]',
            redirectUrl: window.location.origin,
            postLogoutRedirectUri: window.location.origin,
            clientId: 'Portal',
            scope: 'openid email profile offline_access roles internal_api external_api', // 'openid profile offline_access ' + your scopes
            responseType: 'code',
            silentRenew: true,
            useRefreshToken: true,
            renewTimeBeforeTokenExpiresInSeconds: 30,
            ignoreNonceAfterRefresh: true,
            logLevel: LogLevel.Debug
        }
```

TarasKovalenko:
@medeirosrich I have the same configuration and everything works well with multiple tabs.

medeirosrich:
I don't know if this helps, but here is a side by side of the logs from the two tabs that are colliding

[...]

PR #2237, opened 2026-07-03
feat: add useRefreshTokenLock to serialize refresh token renewal across tabs

### LINKED ISSUE #1838: [Bug]: Silent refresh fails using localStorage when multiple tabs trying to refresh at the same time
### Version

16.0.0

_No response_

```true
myConfigId - silent renew failed! Error: Error: authorizedCallback, token(s) validation failed, resetting.
```

```true
1) Set up OIDC Code Flow PKCE using refresh tokens (used 'ng add angular-auth-oidc-client')

2) Set up custom storage using localStorage (used example from https://angular-auth-oidc-client.com/docs/documentation/custom-storage)

3) To speed up testing, setup your authority to have short lived access and refresh tokens. (i.e. 30 - 50 seconds) 

4) Login using one browser [chrome] tab (i.e. this.oidcSecurityService.authorize()). Open console and verify refresh is working

5) Open a few more tabs. Open console and verify they are authorized and refreshing (in my case i open 6 or more)

6) Wait until error occurs. It shouldn't take long - usually 2-20 minutes with 30 - 40 seconds tokens. Open more tabs if needed.
```

```true
Access token should refresh successfully and all tabs remain authorized
```

### Additional context

The same problem was described in [#1662](https://github.com/damienbod/angular-auth-oidc-client/issues/1662)

image

A temporary solution does not suit me well, because my token functions for 300 seconds.

My config: 
image

AuthConfigModule added to the imports of the main AppModule.

RhisiartK:
## Summary
Closes #1838 (same issue previously reported in #1662). New opt-in `useRefreshTokenLock` (default `false`): refresh token renewal is serialised across browser tabs per configId via the Web Locks API.

## The bug
With multiple tabs sharing storage (e.g., a custom localStorage `AbstractSecurityStorage`) and refresh token rotation, tabs renew concurrently using the same one-time-use refresh token. The losing tabs fail validation:

```
silent renew failed! Error: Error: authorizedCallback, token(s) validation failed, resetting.
```

And the session is reset for every tab. Reported with full repro steps in #1662 and #1838.

## The fix
With `useRefreshTokenLock: true`, `refreshSessionWithRefreshTokens` wraps the refresh in `navigator.locks.request('angular-auth-oidc-client-refresh-token-<configId>', ...)`:

- The tab snapshots the access token before requesting the lock and re-reads it once the lock is granted.
- Token changed while waiting and stored tokens are valid → another tab already refreshed; the stored tokens are reused, and the token request is skipped.
- Token changed but no longer valid → refresh normally (the Web Locks spec puts no upper bound on the wait, e.g. suspend/resume).
- Token unchanged → this tab performs the refresh.
- `navigator.locks` unavailable → the refresh runs unchanged without the lock.
- Errors inside the lock follow the existing reset + stop-periodic-check path.

Same approach as angular-oauth2-oidc: https://github.com/manfredsteyer/angular-oauth2-oidc/pull/1423

## Test plan
- [x] New tests: token reuse, refresh-when-unchanged, refresh-when-tokens-invalid, per-config lock name, no-API fallback, error propagation inside the lock
- [x] Full lib suite passes (1040 tests)
- [x] lint / prettier / block-words hooks pass

### DIFF
--- a/projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
@@ -1,5 +1,5 @@
 import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
-import { firstValueFrom, of, throwError } from 'rxjs';
+import { firstValueFrom, NEVER, of, throwError } from 'rxjs';
 import { mockProvider } from '../../test/auto-mock';
 import { AuthStateService } from '../auth-state/auth-state.service';
 import { CallbackContext } from '../flows/callback-context';
@@ -338,6 +338,35 @@ describe('RefreshSessionRefreshTokenService', () => {
           firstValueFrom(
             refreshSessionRefreshTokenService.refreshSessionWithRefreshTokens(
               { configId: 'configId1', useRefreshTokenLock: true },
+              [{ configId: 'configId1' }]
+            )
+          )
+        ).toBeRejected();
+
+        expect(resetAuthorizationDataSpy).toHaveBeenCalled();
+      });
+
+      it('resetAuthorizationData when the refresh exceeds silentRenewTimeoutInSeconds inside the lock', async () => {
+        Object.defineProperty(navigator, 'locks', {
+          value: {
+            request: (_name: string, cb: () => Promise<unknown>) => cb(),
+          },
+          configurable: true,
+        });
+        spyOn(flowsService, 'processRefreshToken').and.returnValue(NEVER);
+        const resetAuthorizationDataSpy = spyOn(
+          resetAuthDataService,
+          'resetAuthorizationData'
+        );
+
+        await expectAsync(
+          firstValueFrom(
+            refreshSessionRefreshTokenService.refreshSessionWithRefreshTokens(
+              {
+                configId: 'configId1',
+                useRefreshTokenLock: true,
+                silentRenewTimeoutInSeconds: 0.01,
+              },
               [{ configId: 'configId1' }]
             )
           )
--- a/projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
@@ -109,6 +109,31 @@ describe('RefreshSessionRefreshTokenService', () => {
     }));
 
     describe('cross-tab refresh token lock', () => {
+      it('does not request a lock when useRefreshTokenLock is disabled', async () => {
+        const requestSpy = jasmine
+          .createSpy('request')
+          .and.callFake((_name: string, cb: () => Promise<unknown>) => cb());
+
+        Object.defineProperty(navigator, 'locks', {
+          value: { request: requestSpy },
+          configurable: true,
+        });
+        const processSpy = spyOn(
+          flowsService,
+          'processRefreshToken'
+        ).and.returnValue(of({} as CallbackContext));
+
+        await firstValueFrom(
+          refreshSessionRefreshTokenService.refreshSessionWithRefreshTokens(
+            { configId: 'configId1', useRefreshTokenLock: false },
+            [{ configId: 'configId1' }]
+          )
+        );
+
+        expect(requestSpy).not.toHaveBeenCalled();
+        expect(processSpy).toHaveBeenCalled();
+      });
+
       it('reuses the stored tokens and skips processRefreshToken when another tab already refreshed', async () => {
         Object.defineProperty(navigator, 'locks', {
           value: {
--- a/projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
+++ b/projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
@@ -1,6 +1,7 @@
 import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
-import { of, throwError } from 'rxjs';
+import { firstValueFrom, of, throwError } from 'rxjs';
 import { mockProvider } from '../../test/auto-mock';
+import { AuthStateService } from '../auth-state/auth-state.service';
 import { CallbackContext } from '../flows/callback-context';
 import { FlowsService } from '../flows/flows.service';
 import { ResetAuthDataService } from '../flows/reset-auth-data.service';
@@ -13,6 +14,7 @@ describe('RefreshSessionRefreshTokenService', () => {
   let intervalService: IntervalService;
   let resetAuthDataService: ResetAuthDataService;
   let flowsService: FlowsService;
+  let authStateService: AuthStateService;
 
   beforeEach(() => {
     TestBed.configureTestingModule({
@@ -23,6 +25,7 @@ describe('RefreshSessionRefreshTokenService', () => {
         mockProvider(FlowsService),
         mockProvider(ResetAuthDataService),
         mockProvider(IntervalService),
+        mockProvider(AuthStateService),
       ],
     });
   });
@@ -34,6 +37,12 @@ describe('RefreshSessionRefreshTokenService', () => {
     );
     intervalService = TestBed.inject(IntervalService);
     resetAuthDataService = TestBed.inject(ResetAuthDataService);
+    authStateService = TestBed.inject(AuthStateService);
+  });
+
+  afterEach(() => {
+    // cleanup the navigator.locks shadow defined by the lock tests
+    delete (navigator as any).locks;
   });
 
   it('should create', () => {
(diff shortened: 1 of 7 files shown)


projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
```diff
@@ -97,5 +106,173 @@ describe('RefreshSessionRefreshTokenService', () => {
       tick();
```
pixelbits-mk, line 107: 🔵 **Informational**: No test for useRefreshTokenLock: false while navigator.locks is present

All lock tests set useRefreshTokenLock to true. None confirm that with it false (or unset), processRefreshToken still runs directly even though real browser test runners expose navigator.locks.

**Recommendation**: Add a test with navigator.locks stubbed and useRefreshTokenLock: false, asserting navigator.locks.request is never called and processRefreshToken runs directly.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
```diff
@@ -22,20 +24,81 @@ export class RefreshSessionRefreshTokenService {
...
+  private refreshWithLock(
+    config: OpenIdConfiguration,
+    allConfigs: OpenIdConfiguration[],
+    customParamsRefresh?: { [key: string]: string | number | boolean }
+  ): Observable<CallbackContext> {
+    const lockName = `angular-auth-oidc-client-refresh-token-${config.configId}`;
+
+    return defer(async (): Promise<CallbackContext> => {
+      const accessTokenBeforeLock =
+        this.authStateService.getAccessToken(config);
```
pixelbits-mk, line 64: 🔵 **Informational**: No test for wasRefreshedInAnotherTab when no access token existed before the lock

getAccessToken can legitimately return null before the very first refresh. Existing tests only cover old-vs-new non-empty tokens, not a falsy-before/truthy-after transition.

**Recommendation**: Add a test where getAccessToken returns undefined before the lock and a real token after, confirming a normal refresh happens rather than a false reuse match.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
```diff
@@ -22,20 +24,81 @@ export class RefreshSessionRefreshTokenService {
...
+  }
+
+  private refreshWithLock(
+    config: OpenIdConfiguration,
+    allConfigs: OpenIdConfiguration[],
+    customParamsRefresh?: { [key: string]: string | number | boolean }
+  ): Observable<CallbackContext> {
+    const lockName = `angular-auth-oidc-client-refresh-token-${config.configId}`;
+
+    return defer(async (): Promise<CallbackContext> => {
```
pixelbits-mk, line 62: 🟡 **Consider Changing**: Refresh-token lock has no timeout, so a stuck request can stall renewal in every tab

navigator.locks.request wraps processRefreshToken with no timeout or AbortSignal. A slow or hung STS call would block every other tab sharing the lock indefinitely. Is this an acceptable tradeoff given silentRenewTimeoutInSeconds isn't wired in?

**Recommendation**: Pass an AbortSignal (e.g. AbortSignal.timeout based on silentRenewTimeoutInSeconds) to navigator.locks.request so a stalled refresh can't block other tabs forever.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
```diff
@@ -1,6 +1,7 @@
 import { inject, Injectable } from '@angular/core';
-import { Observable, throwError } from 'rxjs';
+import { defer, firstValueFrom, Observable, throwError } from 'rxjs';
```
pixelbits-mk: 🔴 **Must Fix**: firstValueFrom import may break declared rxjs 6 peer-dependency support

firstValueFrom was added in rxjs 7, yet peerDependencies reportedly still allow rxjs ^6.5.3. This is the only production file using it, so rxjs-6 consumers could hit a missing export.

**Recommendation**: Replace firstValueFrom with an rxjs-6-compatible subscription pattern, or bump the peer dependency floor to rxjs 7 and document the breaking change.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
```diff
@@ -22,20 +24,81 @@ export class RefreshSessionRefreshTokenService {
...
+            currentAccessToken !== accessTokenBeforeLock &&
+            this.authStateService.areAuthStorageTokensValid(config);
+
+          if (wasRefreshedInAnotherTab) {
+            this.loggerService.logDebug(
+              config,
+              'access token was already refreshed in another tab, reusing the stored tokens'
+            );
+
+            return {
```
pixelbits-mk: 🔵 **Informational**: Hand-crafted CallbackContext duplicates the shape built elsewhere

This object literal duplicates the CallbackContext shape already built in refresh-session-callback-handler.service.ts, hardcoding fields like code: '' and state: ''. The two constructions could drift apart later.

**Recommendation**: Extract a small factory for building a CallbackContext from stored tokens so both call sites stay in sync.

pixelbits-mk: 🔴 **Must Fix**: Lock-reuse path skips the normal callback pipeline, so state and events never publish

When another tab already refreshed, the code hand-builds a CallbackContext, bypassing stateValidationCallbackHandlerService and userHandlerService. authStateService.setAuthorizationData never fires, so authenticated$/NewAuthenticationResult/user data miss the renewal in this tab. Is it fine for app code to simply miss a background refresh a sibling tab performed?

**Recommendation**: Route the reuse case through the same callback stages (or an equivalent 'republish current state' call) so consumers get consistent events regardless of which tab refreshed.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
```diff
@@ -22,20 +24,81 @@ export class RefreshSessionRefreshTokenService {
   ): Observable<CallbackContext> {
     this.loggerService.logDebug(config, 'BEGIN refresh session Authorize');
     let refreshTokenFailed = false;
+    const useLock =
```
pixelbits-mk: 🔵 **Informational**: navigator accessed directly instead of via the DOCUMENT-style DI pattern used elsewhere

The service reads navigator/navigator.locks off the global object, unlike other browser globals in this codebase that use inject(DOCUMENT). Tests monkey-patch navigator instead of using DI mocks.

**Recommendation**: Introduce an injectable wrapper/token for navigator.locks so it can be mocked via TestBed providers instead of patching the global object.

pixelbits-mk (approved):
## 👍 APPROVED WITH COMMENTS

**Found**: 2 Must Fix, 1 Consider Changing, 4 Informational
*Tiers by severity: critical/high → 🔴 Must Fix · medium → 🟡 Consider Changing · low/info → 🔵 Informational*
*Decided by policy rules: `high-architecture-review` · `judgment-needs-human` · `conflicts-need-human` · `reviewer-advises-feedback` · `manual-publish-requested`*

### What this PR does

This PR adds an opt-in useRefreshTokenLock flag that serializes refresh-token renewal across tabs via the Web Locks API, fixing the multi-tab race from #1662/#1838. The new refreshWithLock path snapshots the access token, waits on a per-config lock, and reuses stored tokens if a sibling tab already refreshed. Docs, default config, and six new spec cases line up cleanly with the description. The lock-reuse fast path, though, bypasses the normal callback pipeline, so this tab's own auth-state events may never fire on a renewal done elsewhere.

Reviewed by: 🧪 QA · 🏗️ Architecture · 👨‍💻 Developer · 🔒 Security

*This touches token refresh logic and cross-tab coordination via a new lock path, which is security- and correctness-sensitive (token reuse, race handling, error fallback). Tests are included but the concurrency logic and fallback behavior warrant careful review of edge cases and design fit.*

### Agent feedback

- 🏗️ **Architecture** — Flags that the lock-reuse fast path skips the state-validation/user-data pipeline so this tab's events don't fire, and notes navigator is used directly instead of via DI.
- 🔒 **Security** — Likes the overall design but points out the lock request has no timeout, so a stuck refresh in one tab could block renewal in every other tab.
- 👨‍💻 **Developer** — Raises a likely rxjs 6 peer-dependency break from importing firstValueFrom, plus the same missing-pipeline concern on the reuse path and some minor duplication.
- 🧪 **QA** — Coverage of the six test-plan scenarios looks solid but is missing tests for useRefreshTokenLock: false gating and a null access token before the first refresh.

### 🔴 Must Fix

| # | Finding | Recommendation |
|---|---|---|
| 1 | firstValueFrom import may break declared rxjs 6 peer-dependency support<br>`projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts:2` | Replace firstValueFrom with an rxjs-6-compatible subscription pattern, or bump the peer dependency floor to rxjs 7 and document the breaking change. |
| 2 | Lock-reuse path skips the normal callback pipeline, so state and events never publish<br>`projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts:80` | Route the reuse case through the same callback stages (or an equivalent 'republish current state' call) so consumers get consistent events regardless of which tab refreshed. |

### Consider Changing & Informational

| # |  | Finding | Recommendation |
|---|---|---|---|
| 1 | 🟡 | Refresh-token lock has no timeout, so a stuck request can stall renewal in every tab<br>`projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts:60` | Pass an AbortSignal (e.g. AbortSignal.timeout based on silentRenewTimeoutInSeconds) to navigator.locks.request so a stalled refresh can't block other tabs forever. |
| 2 | 🔵 | No test for useRefreshTokenLock: false while navigator.locks is present<br>`projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts:106` | Add a test with navigator.locks stubbed and useRefreshTokenLock: false, asserting navigator.locks.request is never called and processRefreshToken runs directly. |
| 3 | 🔵 | No test for wasRefreshedInAnotherTab when no access token existed before the lock<br>`projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts:62` | Add a test where getAccessToken returns undefined before the lock and a real token after, confirming a normal refresh happens rather than a false reuse match. |
| 4 | 🔵 | Hand-crafted CallbackContext duplicates the shape built elsewhere<br>`projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts:80` | Extract a small factory for building a CallbackContext from stored tokens so both call sites stay in sync. |
| 5 | 🔵 | navigator accessed directly instead of via the DOCUMENT-style DI pattern used elsewhere<br>`projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts:27` | Introduce an injectable wrapper/token for navigator.locks so it can be mocked via TestBed providers instead of patching the global object. |

### ✅ What checks out

The six test-plan bullets (reuse, refresh-when-unchanged, invalid-tokens, per-config lock name, no-API fallback, error propagation) each have a matching it() in the new 'cross-tab refresh token lock' describe block, and useRefreshTokenLock: false is added consistently to both default-config.ts and openid-configuration.ts.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.spec.ts
RhisiartK: Added in dc0affa5 — locks stubbed with a spy, flag off, asserting `request` is never called and `processRefreshToken` runs.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
RhisiartK: Test added in d48ba460. It pins the current behavior for that scenario — reuse: the path is gated on `areAuthStorageTokensValid`, and the tokens reused are the ones the other tab just stored.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
RhisiartK: Fixed in da0d3d34: the refresh running inside the lock is now bounded by `timeout(silentRenewTimeoutInSeconds * 1000)` — the same 'probably stuck' bound `isSilentRenewRunning` uses. On timeout the HTTP request is aborted, the callback promise settles, the lock is released, and the next tab performs the refresh.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
RhisiartK: Fixed in 490fce64 — subscription-based promise instead, no rxjs 7 API left in production code; the peer range stays `^6.5.3 || ^7.4.0`.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
RhisiartK: Fixed in 4a502754 — `createRenewCallbackContext` in `callback-context.helper.ts`, used by both call sites (`state`/`authResult` are the two variable fields).

RhisiartK: Fixed in b9a96445 with the 'republish current state' option: the reuse branch fires `setAuthenticatedAndFireEvent` and `updateAndPublishAuthState` (ValidationResult.Ok), so subscribers in the reusing tab observe the renewal the same way as in the tab that performed it.

projects/angular-auth-oidc-client/src/lib/callback/refresh-session-refresh-token.service.ts
RhisiartK: Fixed in a7f58b63 using the existing `inject(DOCUMENT)` pattern (`document.defaultView?.navigator?.locks`) rather than a new token — consistent with how the lib accesses other browser globals.

joewIST:
We are still experiencing silent renew failures after upgrading to the latest version ans setting useRefreshTokenLock to true:

"silent renew failed! Error: Error: authorizedCallback, token(s) validation failed, resetting. Hash: "

I will need to revert back to our custom solution so token refresh works across tabs again..

PR #2281, opened 2026-08-25
chore: remove unused deps

timdeschryver:
A leftover from the v22 upgrade.
angular/animations is deprecated, so I removed it. While doing so I also came across other packages that are unused.


projects/angular-auth-oidc-client/src/lib/storage/browser-storage.service.spec.ts
```diff
@@ -243,9 +243,9 @@ describe('BrowserStorageService', () => {
 
   describe('hasStorage', () => {
     it('returns false if there is no storage', () => {
-      (Storage as any) = undefined;
+      vi.stubGlobal('Storage', undefined);
```
timdeschryver, line 254: @damienbod this might have been the cause why some test were failing. The previous CI run also failed.

PR #2283, opened 2026-09-01
fix: apply silent renew timeout to the complete refresh token flow

ThorstenKunz:
## Summary

This PR applies the existing `silentRenewTimeoutInSeconds` consistently to the complete refresh-token flow and prevents transient network failures from clearing valid authorization data.

## Problem

Network errors from the refresh-token request are retried by `retryWhen`. Previously, `silentRenewTimeoutInSeconds` only bounded the refresh operation inside the optional Web Locks path. Without `useRefreshTokenLock`, `forceRefreshSession()` could therefore remain pending indefinitely while the token endpoint was unavailable.

The refresh session also cleared stored authorization data for every failure, including temporary network errors and timeouts.

## Changes

- Apply `silentRenewTimeoutInSeconds` to the complete refresh-token branch of `forceRefreshSession()`, including the preceding well-known endpoint lookup.
- Apply the timeout consistently with and without `useRefreshTokenLock`.
- Preserve stored authorization data for network errors and refresh timeouts.
- Keep the existing reset behavior for permanent authentication and token-processing failures.
- Reuse the existing configuration without introducing another public option.

The timeout inside the Web Locks callback is retained so that the pending lock operation settles and releases the lock reliably.

## Tests

Regression tests cover:

- the configured timeout for the complete refresh-token flow;
- the default 20-second timeout;
- timeout behavior with and without Web Locks;
- preservation of authorization data after network failures;
- preservation of authorization data after timeouts;
- the existing reset behavior for permanent errors.

Validation:

- `npm run lint-lib`
- `npm run test-lib-ci` — 73 test files, 1,052 tests
- `npm run build-lib-prod`
- `npm run check-blockwords`
- Coveralls: 95.717%, increased by 0.009%

### DIFF
--- a/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
+++ b/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
@@ -456,6 +456,7 @@ Makes it possible to turn the `iss` validation off per configuration. **You shou
 
 ### `strictIssuerValidationOnWellKnownRetrievalOff`
 
+
 - Type: `boolean`
 - Required: `false`
 
@@ -557,15 +558,6 @@ Activates Pushed Authorisation Requests (PAR) for login and popup login. <br/>
 Controls the periodic retry time interval for retrieving new tokens in seconds. <br/>
 Default = _3_. <br/>
 `silentRenewTimeoutInSeconds` and `tokenRefreshInSeconds` are the upper bounds for this value.
-
-### `refreshTokenMaxRetries`
-
-- Type: `number`
-- Required: `false`
-
-Limits the number of retries after a refresh token request fails because of a network error. <br/>
-By default, network errors are retried until `silentRenewTimeoutInSeconds` is reached. <br/>
-Set this to `0` to disable retries.
 
 ### `ngswBypass`
 
--- a/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
+++ b/docs/site/angular-auth-oidc-client/docs/documentation/configuration.md
@@ -456,7 +456,6 @@ Makes it possible to turn the `iss` validation off per configuration. **You shou
 
 ### `strictIssuerValidationOnWellKnownRetrievalOff`
 
-
 - Type: `boolean`
 - Required: `false`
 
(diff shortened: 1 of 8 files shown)


damienbod:
Why it this required ? We already have the silentRenewTimeoutInSeconds

ThorstenKunz:
> Why it this required ? We already have the silentRenewTimeoutInSeconds

The original reason for adding `refreshTokenMaxRetries` was that the network-error `retryWhen` in `RefreshTokenCallbackHandlerService` could retry indefinitely. `silentRenewTimeoutInSeconds` was only applied inside the Web Locks path, so a normal refresh-token flow without `useRefreshTokenLock` was not bounded by it.

The retry limit additionally provides a deterministic upper bound on the number of token endpoint POSTs, independently of the retry interval. However, I agree that this introduces another configuration option which may not be necessary once `silentRenewTimeoutInSeconds` is consistently applied to the complete refresh-token flow.

I can simplify the PR by removing `refreshTokenMaxRetries` and relying on the existing `silentRenewTimeoutInSeconds` instead. The PR would then focus on:

- applying `silentRenewTimeoutInSeconds` to the complete refresh-token flow, with and without Web Locks;
- preserving the original network error;
- not clearing authentication data for transient network errors or timeouts.

Would you prefer this simplified approach?

damienbod:
This would be good I think to keep it simple without another configuration

ThorstenKunz:
Thanks. I have simplified the PR as suggested. `refreshTokenMaxRetries` and its documentation have been removed completely. The implementation now relies solely on the existing `silentRenewTimeoutInSeconds`, applied to the complete refresh-token flow with and without Web Locks. Transient network errors and timeouts no longer clear the existing authorization data, while permanent authentication and token-processing errors retain the previous reset behavior. The library tests, lint, production build, and Coveralls check are passing.

ThorstenKunz:
One CI note: the library build, lint, tests, production build, and Coveralls check are passing. The three Angular consumer jobs all fail in the identical `Create Angular Project` step (`ng new`), before the PR artifact is installed or tested. This therefore appears unrelated to the PR changes, likely involving the globally installed latest Angular CLI. I cannot rerun the upstream workflow jobs without repository admin permission. Could a maintainer please rerun the failed jobs?

FabianGosebrink:
Hey, thanks for your PR. Can you merge main again in your branch? This should fix the pipeline. Thanks!

PR #2294, opened 2026-09-18
feat(silent-renew): surface the error and the `error_description` of a failed iframe silent renew

sparlampe:
## What

When the silent renew iframe comes back with an OAuth error, `forceRefreshSession()` resolves with a `LoginResponse` whose `errorMessage` is always `''`, and the `error_description` of the callback is dropped entirely. This PR carries both out of the iframe callback:

- `LoginResponse.errorMessage` — the error of the failed silent renew, i.e. the `error` parameter of the callback url such as `login_required`
- `LoginResponse.errorDescription` — **new, optional** — the `error_description` of that callback, when the identity provider sent one

```ts
this.oidcSecurityService
  .forceRefreshSession()
  .subscribe(({ isAuthenticated, errorMessage, errorDescription }) => {
    if (!isAuthenticated && errorDescription === 'the_reason_i_can_recover_from') {
      // re-issue the request once instead of falling back to a full login
    }
  });
```

## Why

Today the only signal a caller gets from a failed silent renew is `isAuthenticated: false`. Whether the session expired, the provider wants an interactive login, or the provider refused for a reason it explains, the app cannot tell — so the only possible reaction is a full redirect login.

Providers do distinguish. An OAuth error response carries `error` plus an optional `error_description` (RFC 6749, 4.1.2.1), and providers use the description to say why the login was refused. Concretely, a Keycloak extension can answer a silent renew with `error=login_required&error_description=<reason>`, where one reason is recoverable by re-issuing the request once while every other `login_required` needs the full login. Only the description tells the two apart.

The error of the iframe callback reaches the app in no other way: `silentRenewEventHandler` subscribes to the callback itself and swallows the error after logging it, and `EventTypes.SilentRenewFailed` only covers errors from *starting* a renew, not from its callback. The completion payload of the iframe is the only route out.

## How

- `SilentRenewService.codeFlowCallbackSilentRenewIframe` throws a `SilentRenewCallbackError`, which keeps the `error_description` of the callback url next to the `error`. The class does not override `name`, so the thrown error still prints and compares as the plain `Error` thrown before — it is only a carrier for the description.
- `silentRenewEventHandler` puts the error message — and the description, when there is one — on the `refreshSessionWithIFrameCompleted$` payload instead of dropping them.
- `RefreshSessionService.forceRefreshSession` maps both onto the returned `LoginResponse`.
- `LoginResponse.errorDescription?: string` is documented on the interface, in the `forceRefreshSession` JSDoc and on the public API docs page.


damienbod (approved):
LGTM

sparlampe:
@damienbod Sorry, I re-requested your review by mistake (misclick). Please ignore!

